PR #45, opened 2021-09-14
Tidy tokens module

adamchainz:
Improve order of functions, some names, and document each.

### DIFF
--- a/src/django_upgrade/fixers/empty_result_set_exception.py
+++ b/src/django_upgrade/fixers/empty_result_set_exception.py
@@ -10,7 +10,7 @@ from tokenize_rt import Offset
 
 from django_upgrade.ast import ast_start_offset
 from django_upgrade.data import Fixer, State, TokenFunc
-from django_upgrade.tokens import rewrite_imports
+from django_upgrade.tokens import update_import_modules
 
 fixer = Fixer(
     __name__,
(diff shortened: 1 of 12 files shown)


adamchainz:
@browniebroke any thoughts?

src/django_upgrade/tokens.py
```diff
@@ -157,13 +199,17 @@ def replace_arguments(
                 raise AssertionError(f"{keyword.arg} argument not found")
 
 
-def update_imports(
+def update_import_names(
```
browniebroke, line 202: This name sounds a lot clearer to me 👍🏻

src/django_upgrade/fixers/empty_result_set_exception.py
```diff
@@ -42,5 +42,5 @@ def visit_ImportFrom(
         and any(alias.name in REWRITES[node.module] for alias in node.names)
     ):
         yield ast_start_offset(node), partial(
-            rewrite_imports, node=node, rewrites=REWRITES
+            update_import_modules, node=node, module_rewrites=REWRITES[node.module]
```
browniebroke, line 45: Ah yes good point, no need to pass the whole `REWRITES` dict 👍🏻

browniebroke (commented):
Nice simplifications, names are a bit clearer.

adamchainz:
Thanks for having a read!

PR #46, opened 2021-09-16
Add a fixer for django.conf.urls deprecations

matthiask:
This fixer doesn't work yet. It somehow doesn't replace `django.conf.urls.url` with `django.urls.re_path`, only with `django.urls.url`.

That's a bit sad. I'm still posting it here. This fixer would make my life so much easier. Thanks :)

### DIFF
--- a/src/django_upgrade/fixers/urls.py
+++ b/src/django_upgrade/fixers/urls.py
@@ -1,0 +1,61 @@
+"""
+Replace django.conf.urls with django.urls
+"""
+import ast
+from functools import partial
+from typing import Iterable, Tuple
+
+from tokenize_rt import Offset
+
+from django_upgrade.ast import ast_start_offset
+from django_upgrade.data import Fixer, State, TokenFunc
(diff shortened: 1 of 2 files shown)


adamchainz:
There's prior art to copy from django-codemod here, which rewrites `url()` calls to `path()` if it can: https://github.com/browniebroke/django-codemod/blob/main/django_codemod/visitors/urls.py

matthiask:
> 
> 
> There's prior art to copy from django-codemod here, which rewrites `url()` calls to `path()` if it can: https://github.com/browniebroke/django-codemod/blob/main/django_codemod/visitors/urls.py

Ah yes, that would be nice too. However, it probbly has a lower priority since `re_path` won't go away soon? (I agree that `path` is nicer but old habits...)

adamchainz:
I don't really want to add a `url` -> `re_path` rewriter when there's the better fixer to adapt. It shouldn't be too much extra work. I even started a branch for this locally already but didn't have time to finish it.

adamchainz:
Superceded by #57, thanks for the initial PR.

PR #57, opened 2021-09-17
Add django.urls fixer

adamchainz:


adamchainz:
Thanks @browniebroke for the django-codemod implementation which I cribbed. At first I was thinking of parsing the path regex with a regex, which would be nuts.

browniebroke:
Amazing @adamchainz! This really makes me think it'll soon be time to put django-codemod into retirement, this tool is becoming a much more promising replacement 😄

matthiask:
Works really well! Thanks 🎉

PR #61, opened 2021-09-18
Add fixer for required on_delete argument

browniebroke:


tests/fixers/test_on_delete.py
```diff
@@ -0,0 +1,99 @@
...
+
+@pytest.mark.parametrize("field_class", ["ForeignKey", "OneToOneField"])
+def test_field_class_imported(field_class):
+    check_noop(
+        f"""\
+        from django.db.models import {field_class}
+        {field_class}("auth.User")
+        """,
+        settings,
+    )
```
browniebroke, line 37: Not supported yet, but maybe this is the 1% case?

tests/fixers/test_on_delete.py
```diff
@@ -0,0 +1,127 @@
...
+)
+@pytest.mark.parametrize("field_class", ["ForeignKey", "OneToOneField"])
+def test_argument_already_set(on_delete_value, field_class):
+    check_noop(
+        f"""\
+        from django.db import models
+        models.{field_class}("auth.User", on_delete={on_delete_value})
+        """,
+        settings,
+    )
```
adamchainz, line 26: Since the existing value doesn't affect our code in any way, I don't think we need to parametrize it

```suggestion
@pytest.mark.parametrize("field_class", ["ForeignKey", "OneToOneField"])
def test_argument_already_set(field_class):
    check_noop(
        f"""\
        from django.db import models
        models.{field_class}("auth.User", on_delete=models.SET_NULL)
        """,
        settings,
    )
```

src/django_upgrade/fixers/on_delete.py
```diff
@@ -0,0 +1,53 @@
...
+) -> Iterable[Tuple[Offset, TokenFunc]]:
+    if (
+        isinstance(node.func, ast.Attribute)
+        and node.func.attr in {"ForeignKey", "OneToOneField"}
+        and "models" in state.from_imports["django.db"]
+        and isinstance(node.func.value, ast.Name)
+        and node.func.value.id == "models"
+    ):
+        if len(node.args) < 2 and all(kw.arg != "on_delete" for kw in node.keywords):
+            yield ast_start_offset(node), partial(add_on_delete_keyword)
```
adamchainz, line 35: No need for `partial` if we aren't binding any args
```suggestion
            yield ast_start_offset(node), add_on_delete_keyword
```

HISTORY.rst
```diff
@@ -12,6 +12,8 @@ History
   Thanks to Bruno Alla for the original implementation of regex-to-path conversion in django-codemod.
   Thanks to Matthias Kestenholz for an initial PR.
 
+* Add fixer for Django 1.9 requirement to pass ``on_delete`` to ``ForeignKey`` and ``OneToOneField``.
```
adamchainz, line 15: thx

```suggestion
* Add fixer for Django 1.9 requirement to pass ``on_delete`` to ``ForeignKey`` and ``OneToOneField``.

  Thanks to Bruno Alla in `PR #61 <https://github.com/adamchainz/django-upgrade/pull/61>`__.
```

README.rst
```diff
@@ -89,6 +89,19 @@ Django 1.9
 
 Based on the `Django 1.9 release notes <https://github.com/django/django/blob/main/docs/releases/1.9.txt>`__.
 
+``on_delete`` argument
+~~~~~~~~~~~~~~~~~~~~~~
+
+Add ``on_delete=models.CASCADE`` to ForeignKey and OneToOneField:
```
adamchainz, line 95:
```suggestion
Add ``on_delete=models.CASCADE`` to ``ForeignKey`` and ``OneToOneField``:
```

adamchainz (commented):
LGTM. One thing is we could insert the argument first, after any positional args, rather than last. This would require passing the number of positional arguments through to `add_on_delete_keyword` It's more informational than other field arguments.

tests/fixers/test_on_delete.py
adamchainz: I think it might be a bit more than 1% of people who like to use this import style. But we can go ahead like so and create an issue to loop back in the future.

browniebroke:
> One thing is we could insert the argument first, after any positional args, rather than last. This would require passing the number of positional arguments through to `add_on_delete_keyword` It's more informational than other field arguments.

That's a good point, also `on_delete` is the second positional argument in the implementation, so that matches up well too. I'll try to make it work...

adamchainz:
I made a commit and it looks okay to me, so merging! Thanks!

tests/fixers/test_on_delete.py
adamchainz: Issue: #62.

browniebroke:
Thanks for finishing it up 👍🏻

PR #140, opened 2022-05-03
Remove default_app_config assignments

### LINKED ISSUE #139: Removal of `default_app_config` on Django 3.2+
### Description

Django 3.2 introduced an automatic [app config discovery mechanism](https://docs.djangoproject.com/en/dev/releases/3.2/#automatic-appconfig-discovery). Do you think django-upgrade could remove the deprecated `default_app_config` application configuration variable?

browniebroke:
Fix #139

### DIFF
--- a/src/django_upgrade/fixers/default_app_config.py
+++ b/src/django_upgrade/fixers/default_app_config.py
@@ -12,7 +12,7 @@ from tokenize_rt import Offset, Token
 
 from django_upgrade.ast import ast_start_offset
 from django_upgrade.data import Fixer, State, TokenFunc
-from django_upgrade.tokens import CODE, find_final_token
+from django_upgrade.tokens import LOGICAL_NEWLINE, find
 
 fixer = Fixer(
     __name__,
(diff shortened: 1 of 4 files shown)


src/django_upgrade/fixers/default_app_config.py
```diff
@@ -0,0 +1,41 @@
...
+def visit_Assign(
+    state: State,
+    node: ast.Assign,
+    parent: ast.AST,
+) -> Iterable[tuple[Offset, TokenFunc]]:
+    if (
+        state.filename == "__init__.py"
+        and len(node.targets) == 1
+        and isinstance(node.targets[0], ast.Name)
+        and node.targets[0].id == "default_app_config"
```
adamchainz, line 33: I think we should also check the parent is the module

src/django_upgrade/fixers/default_app_config.py
```diff
@@ -0,0 +1,41 @@
...
+    ):
+        yield ast_start_offset(node), partial(remove_assignment, node=node)
+
+
+def remove_assignment(tokens: list[Token], i: int, *, node: ast.Assign) -> None:
+    new_src = ""
+    j = find_final_token(tokens, i, node=node)
+    tokens[i:j] = [Token(name=CODE, src=new_src)]
```
adamchainz, line 41:
```suggestion
    j = find_final_token(tokens, i, node=node)
    tokens[i:j] = [Token(name=CODE, src="")]
```

PR #144, opened 2022-05-04
Fix bug with default_app_config removal in nested __init__.py

browniebroke:
The current version of `default_app_config` removal implemented in #140 only works if the `__init__.py` is at the top level, not if it's under a subdirectory, which is where it's going to be most of the time.

### DIFF
--- a/src/django_upgrade/data.py
+++ b/src/django_upgrade/data.py
@@ -21,7 +21,7 @@ settings_re = re.compile(r"\bsettings\b")
 
 test_re = re.compile(r"(\b|_)tests?(\b|_)")
 
-dunder_init_re = re.compile(r"(^|[\/])__init__.py$")
+dunder_init_re = re.compile(r"(^|[\\/])__init__\.py$")
 
 
 class State:
--- a/src/django_upgrade/data.py
+++ b/src/django_upgrade/data.py
@@ -21,7 +21,7 @@ settings_re = re.compile(r"\bsettings\b")
 
 test_re = re.compile(r"(\b|_)tests?(\b|_)")
 
-dunder_init_re = re.compile(r"\b__init__\.py$")
+dunder_init_re = re.compile(r"\b__init__.py$")
 
 
 class State:
--- a/src/django_upgrade/data.py
+++ b/src/django_upgrade/data.py
@@ -21,7 +21,7 @@ settings_re = re.compile(r"\bsettings\b")
 
 test_re = re.compile(r"(\b|_)tests?(\b|_)")
 
-dunder_init_re = re.compile(r"\b__init__.py$")
+dunder_init_re = re.compile(r"(^|[\/])__init__.py$")
 
 
 class State:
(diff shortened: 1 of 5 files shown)


src/django_upgrade/data.py
```diff
@@ -21,6 +21,8 @@ def __init__(self, target_version: tuple[int, int]) -> None:
 
 test_re = re.compile(r"(\b|_)tests?(\b|_)")
 
+dunder_init_re = re.compile(r"\b__init__.py$")
```
adamchainz, line 24: I think we should check for a path separator rather a word boundary (which would match weird names like `foo-__init__.py`)

something like:

```suggestion
dunder_init_re = re.compile(r"(^|[\/])__init__.py$")
```

tests/test_data.py
```diff
@@ -46,3 +46,39 @@ def test_looks_like_test_file_false(filename):
...
     )
     assert not state.looks_like_test_file()
+
+
+@pytest.mark.parametrize(
+    ("filename"),
+    (
+        "__init__.py",
+        "package/__init__.py",
+        "project/package/__init__.py",
```
adamchainz, line 56: we should test windows style as well

```suggestion
        "__init__.py",
        "package/__init__.py",
        "package\__init__.py",
        "project/package/__init__.py",
        "project\package\__init__.py",
```

tests/test_data.py
```diff
@@ -46,3 +46,39 @@ def test_looks_like_test_file_false(filename):
...
+        filename=filename,
+        from_imports={},
+    )
+    assert state.looks_like_test_file()
+
+
+@pytest.mark.parametrize(
+    ("filename"),
+    (
+        "__thing__init__.py",
```
adamchainz, line 71: for the word separator change

```suggestion
        "__thing__init__.py",
        "thing-__init__.py",
```

src/django_upgrade/data.py
browniebroke: Oh yes, of course... I think I didn't completely undertsand the regex 🙈 

I can make the changes, but I see you started to pick them up?

tests/test_data.py
```diff
@@ -46,3 +46,42 @@ def test_looks_like_test_file_false(filename):
...
+    ("filename"),
+    (
+        "__init__.py",
+        "package/__init__.py",
+        r"package\__init__.py",
+        "project/package/__init__.py",
+        r"project\package\__init__.py",
+    ),
+)
+def looks_like_dunder_init_file_true(filename):
```
adamchainz, line 61: woops this and the below weren't called `test_`, so they didn't run... 100% test coverage to the rescue though! fixing...

browniebroke: Yes, my mistake... 😞  Nice catch coverage ☂️

tests/test_data.py
```diff
@@ -64,7 +64,7 @@ def test_looks_like_dunder_init_file_true(filename):
         filename=filename,
         from_imports={},
     )
-    assert state.looks_like_test_file()
+    assert state.looks_like_dunder_init_file()
```
browniebroke, line 67: Damn... I was on a roll 😢

adamchainz:
released in 1.6.1 !

tests/test_data.py
adamchainz: no worries this is why we have tools :)

PR #151, opened 2022-05-13
Add PyCQA/bandit to pre-commit

cclauss:
https://github.com/PyCQA/bandit


adamchainz:
No thanks... I keep all my open source projects with the same pre-commit config. I've evaluated bandit and found it would have too many false positives, in general.

PR #167, opened 2022-08-07
Update re path

### LINKED ISSUE #60: Rewrite re_path() calls into path()
For people who did find/replace of `url()` to `re_path()`, we could help by applying the same rewriting logic of `url()` to convert `re_path()` to `path()`

This could get a little awkward with removing `re_path()` if it's no longer needed, tracking if that's needed.

UnknownPlatypus:
Resolves #60 

Let me know if my approach is fine for you.
I had to alter some of the existing implementation to keep track of the rewritten nodes and update the imports.


adamchainz:
Thanks for the contribution. I was considering merging it just now before releasing 1.8.0 with #169 , but held off.

Rewriting URL's to `path` *can* cause issues, in rare cases, because it can change types, as reported in [this comment](https://github.com/adamchainz/django-upgrade/issues/60#issuecomment-972716806). For example, the `int` path converter returns an int, where the equivalent regex would still return a string that the view needed to cast.

I think this rarely causes problems, but I am a little hesitant to do *even more* URL rewriting without some consideration. Perhaps putting all the `re_path` -> `path` rewriting behind a flag... or provide a flag to disable it.

Do you have any thoughts? Hard to make a decision without much data, I would like some more input...

UnknownPlatypus:
> Rewriting URL's to path can cause issues, in rare cases, because it can change types, as reported in https://github.com/adamchainz/django-upgrade/issues/60#issuecomment-972716806.

Yeah, this is an issue but not for every converter (only the `int` and `UUID` converter are problematic), maybe we should be more timid and only rewrites the one we know are safe.

> I think this rarely causes problems

I totally agree, if you decide to use a regex like `r"\d+"` on purpose, you 99.9% of the time want to work with an int and you will eighter cast the value to an int at the beginning of the view or lookup the database and have the ORM handle the type coercion for you.

For the UUID case, I have less experience working with them so I'm not quite sure but I feel like it might be similar.

I tried this patch on a 250k python line of code django repo (which started a while ago with django 1.x) and it did rewrite some regex to use the `int` converter (15) but I had no issue nonetheless because in 70% of the cases It was directly calling the ORM with the value and the other 30% were casting it to an `int` at the beginning of the view.

I'm not quite sure what to do here, surely adding a flag is a solution and would ensure this library is completely safe but on the other end, I really feel like this potential issue wont happened ever.

(I've also checked https://github.com/browniebroke/django-codemod and they don't seem to address this.)

tests/fixers/test_django_urls.py
```diff
@@ -320,9 +323,354 @@ def test_complete():
...
+            path('post/<slug:slug>/', views.post, name='post'),
             re_path(r'^post/(?P<slug>[w-]+)/$', views.post, name='post'),
             re_path(r'^weblog/', include('blog.urls')),
         ]
         """,
         settings,
     )
+
+
+@pytest.mark.parametrize(
```
adamchainz, line 335: deliberately didn't use parameterize to save indentation and ensure they're all named, I'm rewriting this

README.rst
```diff
@@ -194,6 +194,11 @@ Rewrites imports of ``include()`` and ``url()`` from ``django.conf.urls`` to ``d
 .. |new path() syntax| replace:: new ``path()`` syntax
 __ https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax
 
+For some cases, this change alters the type of the arguments passed to the view, from ``str`` to the converted type (e.g. ``int``).
+This is not guaranteed backwards compatible: there is a chance that the view expects a string, rather than the converted type.
+But, pragmatically, it seems 99.9% of views do not require strings, and instead work with either strings or the converted type.
+Thus, you should test affected paths after this fixer makes any changes.
```
adamchainz, line 200: added this paragraph

tests/fixers/test_django_urls.py
UnknownPlatypus: Sorry for the inconvenience, I was unsure about the reason behind your choice of not using parametrize. I should have followed the existing style, mb!

PR #169, opened 2022-08-08
Rewrite django.utils.timezone.utc to datetime.timezone.utc

hramezani:
Fixes https://github.com/adamchainz/django-upgrade/issues/135

PR #171, opened 2022-08-17
Fix PR number of utc fixer entry in HISTORY.rst

hramezani:
Fix for https://github.com/adamchainz/django-upgrade/pull/169#issuecomment-1217558391


adamchainz:
Ah yes!

PR #173, opened 2022-08-24
Add USE_L10N fixer

### LINKED ISSUE #161: Delete `USE_L10N` setting if Django version >= 4.0
### Description

The `USE_L10N` setting is deprecated in Django 4.0: https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-USE_L10N

Figure this would be a good candidate for this project. Happy to work on it if you agree.

johnnymetz:
Closes #161.

### DIFF
--- a/src/django_upgrade/fixers/use_l10n.py
+++ b/src/django_upgrade/fixers/use_l10n.py
@@ -32,6 +32,7 @@ def visit_Assign(
         and node.targets[0].id == "USE_L10N"
         and isinstance(node.value, ast.Constant)
         and node.value.value is True
+        and isinstance(parent, ast.Module)
         and state.looks_like_settings_file()
     ):
         yield ast_start_offset(node), partial(erase_node, node=node)
--- a/src/django_upgrade/fixers/use_l10n.py
+++ b/src/django_upgrade/fixers/use_l10n.py
@@ -8,11 +8,11 @@ import ast
 from functools import partial
 from typing import Iterable
 
-from tokenize_rt import Offset, Token
+from tokenize_rt import Offset
 
 from django_upgrade.ast import ast_start_offset
 from django_upgrade.data import Fixer, State, TokenFunc
-from django_upgrade.tokens import LOGICAL_NEWLINE, find
+from django_upgrade.tokens import erase_node
 
 fixer = Fixer(
     __name__,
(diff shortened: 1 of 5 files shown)


src/django_upgrade/fixers/use_l10n.py
```diff
@@ -0,0 +1,44 @@
...
+        and isinstance(node.targets[0], ast.Name)
+        and node.targets[0].id == "USE_L10N"
+        and isinstance(node.value, ast.Constant)
+        and node.value.value is True
+        and state.looks_like_settings_file()
+    ):
+        yield ast_start_offset(node), partial(remove_assignment, node=node)
+
+
+# TODO: copied from default_app_config.py - should we move this to a common
```
johnnymetz, line 40: Question here

adamchainz: Yeah, put it in `tokens`, under the `# Rewriting functions` header. i think it could also be renamed to `remove_logical_line` ?

It also has a mistake to unod: it takes `node` but doesn't use it, so you can drop the arg.

adamchainz (changes requested):

adamchainz:
Nice work.

src/django_upgrade/fixers/use_l10n.py
adamchainz: We already had `erase_node` which works equivalent, so I've updated both fixers to use that.

tests/fixers/test_use_l10n.py
```diff
@@ -0,0 +1,97 @@
...
+
+def test_ignore_conditional():
+    check_noop(
+        """\
+        if something:
+            USE_L10N = True
+        """,
+        settings,
+        filename="myapp/settings.py",
+    )
```
adamchainz, line 47: Added this case and the appropriate conditional `isinstance(parent, ast.Module)`, because erasing the line when it's the only thing in a block can leave the file broken

Indented settings can occur with e.g. django-configurations, I don't wanna worry about handling such situations (right now)

adamchainz (approved):

johnnymetz:
Looks great! Thanks for the help

johnnymetz:
@adamchainz just curious, how did you commit to my fork's branch?

I'm the maintainer of a new public repo and I'm trying to make changes to someone's PR but I can't figure it out. I'm getting the following error when I try to checkout their fork's branch locally:

```
$ git checkout <contributer>:<branch>
error: pathspec '<contributer>:<branch>' did not match any file(s) known to git
```

adamchainz:
`gh pr checkout 173` https://cli.github.com/manual/gh_pr_checkout

commit and push

PR #181, opened 2022-08-29
Update `from_imports` type hint to use `DefaultDict`

UnknownPlatypus:
Following https://github.com/adamchainz/django-upgrade/pull/167#discussion_r954820379, I think this fixes the typing issue while keeping the more accurate type annotation.


adamchainz:
Thanks!

PR #184, opened 2022-09-06
Add fixer to update BaseCommand.requires_system_checks

browniebroke:
https://docs.djangoproject.com/en/dev/releases/3.2/#deprecated-features-3-2

> Using a boolean value in [`BaseCommand.requires_system_checks`](https://docs.djangoproject.com/en/dev/howto/custom-management-commands/#django.core.management.BaseCommand.requires_system_checks) is deprecated. Use `'__all__'` instead of `True`, and `[]` (an empty list) instead of `False`.


src/django_upgrade/data.py
```diff
@@ -38,6 +38,10 @@ def __init__(
     def looks_like_migrations_file(self) -> bool:
         return "migrations" in self.filename.split(os.path.sep)
 
+    def looks_like_admin_command(self) -> bool:
+        path_parts = self.filename.split(os.path.sep)
+        return "management" in path_parts and "commands" in path_parts
```
adamchainz, line 43: I think to be even surer, we could check for "management" followed by "commands" - perhaps with clever use of `list.index()`

src/django_upgrade/fixers/management_command.py
```diff
@@ -0,0 +1,44 @@
...
+
+fixer = Fixer(
+    __name__,
+    min_version=(3, 2),
+)
+
+REPLACEMENTS = {
+    True: '"__all__"',
+    False: "[]",
+}
```
adamchainz, line 25: I think this would be clearer just as an `if` statement inside the function, there aren't any othre cases that may be added in the future

README.rst
```diff
@@ -481,6 +481,20 @@ Removes module-level ``default_app_config`` assignments from ``__init__.py`` fil
 
     -default_app_config = 'my_app.apps.AppConfig'
 
+``BaseCommand.requires_system_checks``
```
adamchainz, line 484: I think I tried to keep the fixer docs in the same order as the deprecation docs, so this would be before `EmailValidator`

adamchainz (commented):
Nice, nice. This can be a real nuisance in some projects. I'm gonna just make the below little fixes.

adamchainz:
Yeah! 👍👏

Issue #182, opened 2022-08-29
Rewrite `admin.site.register()` to `@admin.register()`


UnknownPlatypus:
### Description
We could rewrite `admin.site.register()` to the more elegant `@admin.register()` syntax :  

```diff
+@admin.register(MyModel)
class MyCustomAdmin:
    pass

-admin.site.register(MyModel, MyCustomAdmin)
```

This syntax was actually added a while ago ([Django1.7](https://docs.djangoproject.com/en/dev/releases/1.7/#minor-features)). 
I usually find it a bit easier to use since the Model and associated Custom Admin definition stay really close. 

 ### Potential Issue
There is a [known issue](https://docs.djangoproject.com/en/dev/ref/contrib/admin/#the-register-decorator) when using a python 2 style super call in the `__init__` method. **The fixer should not rewrite in this case.**
> You can’t use this decorator if you have to reference your model admin class in its \_\_init__() method, e.g. super(PersonAdmin, self).\_\_init__(*args, **kwargs). You can use super().\_\_init__(*args, **kwargs).

Let me know if this is something you'll be interested in, I'll be happy to submit a PR.
Cheers

adamchainz:
This could be an interesting fixer to have, but it would also be more of a coding style choice. The "direct register" approach is still used in Django's examples, and still needed for cases where model classes don't map to admin classes one-to-one.

Also, wouldn't the caveat also extend to other references to the class within the body, e.g. in other `super()` calls? Or is it just `__init__`?

UnknownPlatypus:
Thanks for the quick answer! I agree it's more of a coding style thing (although I do think it makes the navigation easier).

> and still needed for cases where model classes don't map to admin classes one-to-one

I had this point in mind but I think we can deal with it, if a custom admin classe maps to several models, it's either all happening in one file and we can track the number of usage and skip rewrite if it's more than one OR the custom admin class is imported in another file and in that case we skip the rewrite anyway. Will need a few test cases but I think it's manageable.

> Also, wouldn't the caveat also extend to other references to the class within the body, e.g. in other super() calls? Or is it just __init__?

It has to do with the class creation and the way the decorator works, initially was flagged for the `__init__` method (See [initial ticket](https://code.djangoproject.com/ticket/24323) and associated [MR](https://github.com/django/django/pull/5368)) but after testing, the `__new__` dunder method seems to have the same caveat. Other methods seems to be working juste fine.

browniebroke:
[...]

PR #189, opened 2022-09-10
Rewrite `admin.site.register()` calls to `@admin.register()`

UnknownPlatypus:
Implementation of the `admin.site.register` rewriter idea of #182 

The following methodology was used:
- `visit_ClassDef` to track custom admin candidates.
- `visit_Call` to find old style `__init__` / `__new__` methods. If any match, remove the last custom admin class candidate.
- `visit_Call` again to find exact `admin.site.register(MyModel, MyCustomAdmin)` match with `MyCustomAdmin` being one of the remaining candidates. Associated models are stored in a set so that if we have multiple match (`one-to-many` relation), we don't rewrite.

When backtracking, only remaining candidates are rewritten.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -127,27 +127,27 @@ A single ``else`` block may be present, but ``elif`` is not supported.
 See also `pyupgrade’s similar feature <https://github.com/asottile/pyupgrade/#python2-and-old-python3x-blocks>`__ that removes outdated code from checks on the Python version.
 
 Django 1.7
------------
+----------
 
-`Release Notes <https://docs.djangoproject.com/en/dev/releases/1.7/>`__
+`Release Notes <https://docs.djangoproject.com/en/stable/releases/1.7/>`__
 
 Admin model registration
-~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~
 
 Rewrites ``admin.site.register()`` calls to the new ``@admin.register()`` decorator syntax when eligible.
 
 .. code-block:: diff
 
     +@admin.register(MyModel)
-    class MyCustomAdmin:
-        pass
+     class MyCustomAdmin:
+         pass
 
     -admin.site.register(MyModel, MyCustomAdmin)
 
 Django 1.9
 -----------
 
-`Release Notes <https://docs.djangoproject.com/en/dev/releases/1.9/>`__
+`Release Notes <https://docs.djangoproject.com/en/stable/releases/1.9/>`__
 
 ``on_delete`` argument
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -126,10 +126,28 @@ A single ``else`` block may be present, but ``elif`` is not supported.
 
 See also `pyupgrade’s similar feature <https://github.com/asottile/pyupgrade/#python2-and-old-python3x-blocks>`__ that removes outdated code from checks on the Python version.
 
+Django 1.7
+-----------
+
+`Release Notes <https://docs.djangoproject.com/en/dev/releases/1.7/>`__
+
+Admin model registration
+~~~~~~~~~~~~~~~~~~~~~~
+
+Rewrites ``admin.site.register()`` calls to the new ``@admin.register()`` decorator syntax when eligible.
+
+.. code-block:: diff
+
+    +@admin.register(MyModel)
+    class MyCustomAdmin:
+        pass
+
+    -admin.site.register(MyModel, MyCustomAdmin)
+
 Django 1.9
 -----------
 
-`Release Notes <https://github.com/django/django/blob/main/docs/releases/1.9.txt>`__
+`Release Notes <https://docs.djangoproject.com/en/dev/releases/1.9/>`__
 
 ``on_delete`` argument
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/src/django_upgrade/fixers/admin_register.py
+++ b/src/django_upgrade/fixers/admin_register.py
@@ -44,13 +44,13 @@ def visit_ClassDef(
 def update_class_def(
     tokens: list[Token], i: int, *, node: ast.ClassDef, state: State
 ) -> None:
-    arg_kwargs = class_to_decorate.get(state, {}).pop(node.name, set())
-    if len(arg_kwargs) == 1:
+    model_names = class_to_decorate.get(state, {}).pop(node.name, set())
+    if len(model_names) == 1:
         j, indent = extract_indent(tokens, i)
         insert(
             tokens,
             j,
-            new_src=f"{indent}@admin.register({arg_kwargs.pop()})\n",
+            new_src=f"{indent}@admin.register({model_names.pop()})\n",
         )
 
 
@@ -76,7 +76,7 @@ def visit_Call(
         )
     ):
         admin_model_name = node.args[1].id
-        if admin_model_name in class_to_decorate.get(state, {}).keys():
+        if admin_model_name in class_to_decorate.get(state, {}):
             class_to_decorate[state][admin_model_name].add(node.args[0].id)
             yield ast_start_offset(node), partial(
                 erase_register_node,
@@ -104,6 +104,6 @@ def visit_Call(
 def erase_register_node(
     tokens: list[Token], i: int, *, node: ast.Call, admin_model_name: str, state: State
 ) -> None:
-    arg_kwargs = class_to_decorate.get(state, {}).get(admin_model_name, set())
-    if len(arg_kwargs) == 1:
+    model_names = class_to_decorate.get(state, {}).get(admin_model_name, set())
+    if len(model_names) == 1:
         erase_node(tokens, i, node=node)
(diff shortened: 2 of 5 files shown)


src/django_upgrade/fixers/admin_register.py
```diff
@@ -0,0 +1,109 @@
...
+        and node.func.value.value.id == "admin"
+        and not node.keywords
+        and (
+            len(node.args) == 2
+            and isinstance(node.args[0], ast.Name)
+            and isinstance(node.args[1], ast.Name)
+        )
+    ):
+        admin_model_name = node.args[1].id
+        if admin_model_name in class_to_decorate.get(state, {}).keys():
```
adamchainz, line 79: should work without `.keys()` since `in` on a dict is on the keys by default
```suggestion
        if admin_model_name in class_to_decorate.get(state, {}):
```

src/django_upgrade/fixers/admin_register.py
```diff
@@ -0,0 +1,109 @@
...
+def update_class_def(
+    tokens: list[Token], i: int, *, node: ast.ClassDef, state: State
+) -> None:
+    arg_kwargs = class_to_decorate.get(state, {}).pop(node.name, set())
+    if len(arg_kwargs) == 1:
+        j, indent = extract_indent(tokens, i)
+        insert(
+            tokens,
+            j,
+            new_src=f"{indent}@admin.register({arg_kwargs.pop()})\n",
```
adamchainz, line 53: better variable name?
```suggestion
    model_names = class_to_decorate.get(state, {}).pop(node.name, set())
    if len(model_names) == 1:
        j, indent = extract_indent(tokens, i)
        insert(
            tokens,
            j,
            new_src=f"{indent}@admin.register({model_names.pop()})\n",
```

src/django_upgrade/fixers/admin_register.py
```diff
@@ -0,0 +1,109 @@
...
+        and isinstance(parent, ast.Attribute)
+        and parent.attr in {"__init__", "__new__"}
+        and len(node.args) == 2
+        and isinstance(node.args[0], ast.Name)
+        and isinstance(node.args[1], ast.Name)
+    ):
+        # Cannot convert class using py2 style `super(MyAdmin, self)`
+        # in its `__init__` or `__new__` method.
+        # https://docs.djangoproject.com/en/dev/ref/contrib/admin/#the-register-decorator
+        class_to_decorate[state].popitem()
```
adamchainz, line 101: this isn't exactly precise, it's not guaranteed the `__init__` / `__new__` is part of the popped admin class, and special cases like multiple `super()` calls will cause multiple admin classes to be popped, e.g.

```
class HamAdmin(...):
    ...


class SpamAdmin(...):
    def __init__(self, *args, **kwargs):
        if sys.version_info >= (3, 10):
			# something
            super(SpamAdmin, self).__init__(*args, **kwargs)
		else:
			# something else
            super(SpamAdmin, self).__init__(*args, **kwargs)
```

I'd feel more confident if the code checked for the special case in `visit_ClassDef`. It could walk the class def body, looking for `__init__` / `__new__` function defs, and then check their bodies for any unsupported `super()` calls. If there's a match, no need to ever add the class to the to-decorate list.

src/django_upgrade/fixers/admin_register.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def erase_register_node(
+    tokens: list[Token], i: int, *, node: ast.Call, admin_model_name: str, state: State
+) -> None:
+    arg_kwargs = class_to_decorate.get(state, {}).get(admin_model_name, set())
+    if len(arg_kwargs) == 1:
```
adamchainz, line 108:
```suggestion
    model_names = class_to_decorate.get(state, {}).get(admin_model_name, set())
    if len(model_names) == 1:
```

adamchainz (changes requested):
Nice work, some comments but overall very impressive and thought through

WOuld also need docs and a changelog note :)

src/django_upgrade/fixers/admin_register.py
UnknownPlatypus: Absolutely!

src/django_upgrade/fixers/admin_register.py
UnknownPlatypus: Oh you're right! Didn't thought about these cases. Will look into that tomorrow, should not be too difficult

UnknownPlatypus: I've moved the implementation inside the `visit_ClassDef`, this is not perfect because it's walking the tree another time but only when specific conditions are met so it's probably fine.

adamchainz: Yeah I think the overhead should be fine.

README.rst
```diff
@@ -126,10 +126,28 @@ A single ``else`` block may be present, but ``elif`` is not supported.
 
 See also `pyupgrade’s similar feature <https://github.com/asottile/pyupgrade/#python2-and-old-python3x-blocks>`__ that removes outdated code from checks on the Python version.
 
+Django 1.7
+-----------
+
+`Release Notes <https://docs.djangoproject.com/en/dev/releases/1.7/>`__
```
adamchainz, line 132: slightly better to link to `en/stable/releases` so the big yellow "unsupported version" banner isn't present

README.rst
```diff
@@ -126,10 +126,28 @@ A single ``else`` block may be present, but ``elif`` is not supported.
...
+    class MyCustomAdmin:
+        pass
+
+    -admin.site.register(MyModel, MyCustomAdmin)
+
 Django 1.9
 -----------
 
-`Release Notes <https://github.com/django/django/blob/main/docs/releases/1.9.txt>`__
+`Release Notes <https://docs.djangoproject.com/en/dev/releases/1.9/>`__
```
adamchainz, line 150: yes nice spot

README.rst
```diff
@@ -126,10 +126,28 @@ A single ``else`` block may be present, but ``elif`` is not supported.
 
 See also `pyupgrade’s similar feature <https://github.com/asottile/pyupgrade/#python2-and-old-python3x-blocks>`__ that removes outdated code from checks on the Python version.
 
+Django 1.7
+-----------
```
adamchainz, line 130: FYI, in reST the headers should match the text length exactly, I think if they don't match it can break PyPI

I should add a pre-commit check for this

anyway, I'm fixing

src/django_upgrade/fixers/admin_register.py
```diff
@@ -0,0 +1,115 @@
...
+
+from tokenize_rt import Offset, Token
+
+from django_upgrade.ast import ast_start_offset
+from django_upgrade.data import Fixer, State, TokenFunc
+from django_upgrade.tokens import erase_node, extract_indent, insert
+
+fixer = Fixer(
+    __name__,
+    min_version=(1, 7),
```
adamchainz, line 20: added 1.7 and 1.8 to `--target-version` options so that it makes sense to have a 1.7+ fixer

src/django_upgrade/fixers/admin_register.py
```diff
@@ -0,0 +1,115 @@
...
+                        isinstance(target_node, ast.Attribute)
+                        and target_node.attr in {"__init__", "__new__"}
+                        and isinstance(target_node.value, ast.Call)
+                        and isinstance(target_node.value.func, ast.Name)
+                        and target_node.value.func.id == "super"
+                        and len(target_node.value.args) == 2
+                        and isinstance(target_node.value.args[0], ast.Name)
+                        and isinstance(target_node.value.args[1], ast.Name)
+                    ):
+                        return
```
adamchainz, line 55: Let's put this in a different function

adamchainz (commented):
I came up with two `super()` defeating cases, and then I fixed them by reducing the logic around checking for the problem. Basically the original code was too specific, when we need to more generic just in case there's code with weird patterns out there.

The previous version checked for the full syntax structure `super(X, y).__init__(...)`, in exactly an `__init__` function that was directly in the class body. I reduced that to "any sub-node that's a function that calls super with two args at some point".

I also found this was a good time to use the `NodeVisitor` class, since we can avoid some of the overhead of `ast.walk()` by avoiding descending into the `FunctionDef`s if they don't have the right name.

adamchainz (approved):

README.rst
UnknownPlatypus: Oh great, I was looking for something like that. Good to know!

README.rst
UnknownPlatypus: Interesting, sry for that. I'm not familiar enough with the rst syntax. Kinda scary to be able to have trouble with pypi from that

UnknownPlatypus:
Looking really great! Thanks for the review and nice catch about the 2 other edge cases. Quite unusual cases I wouldn't have thought about (and didn't faced while testing).

README.rst
adamchainz: If you ever see a package on PyPI whose README looks like rst syntax rather than rendered, now you know

I use `twine check` at release time which catches this: https://twine.readthedocs.io/en/stable/#twine-check . But it would be good to have it caught on CI. One day I'll figure it out.

adamchainz:
Yeah they must be really rare, but when writing an automated refactor it's best to be as defensive as possible.

cclauss:
I am new to Django.

This change converts `-admin.site.register(ApiConsumer, ApiConsumerAdmin)`
to `+@admin.register(ApiConsumer)`

Is it OK that there is no `+@admin.register(ApiConsumerAdmin)` decorator?

Issue #190, opened 2022-09-12
Extensions to `@admin.register` fixer


adamchainz:
### Description

Following #182 / #189 , there are a few more cases that could be covered with further work.

1. Support kwarg form ✅ #192

```diff
+@admin.register(MyModel)
 class MyModelAdmin(...):
    ...

-admin.site.register(MyModel, admin_class=MyModelAdmin)
```

2. Support multiple models ✅ #200

```diff
+@admin.register(Model1, Model2)
 class MyModelAdmin(...):
    ...

-admin.site.register(Model1, MyModelAdmin)
-admin.site.register(Model2, MyModelAdmin)
```

```diff
+@admin.register(Model1, Model2)
 class MyModelAdmin(...):
    ...

-admin.site.register((Model1, Model2), MyModelAdmin)
```

3. Support when there are other class decorators, insert at top ✅ #193

```diff
+@admin.register(MyModel)
 @something
 class MyModelAdmin(...):
    ...

-admin.site.register(MyModel, MyModelAdmin)
```

4. Support custom admin sites ✅ #228

```diff
from myapp.admin import custom_site

+@admin.register(MyModel, site=custom_site)
 class MyModelAdmin(...):
    ...

-custom_site.register(MyModel, MyModelAdmin)
```

- detect custom site objects heuristically, to avoid affecting other code that uses a 'register' pattern, perhaps when the object's name ends in 'site', and the registered class name ends in 'Admin' (and maybe the file looks like an admin file, called "admin.py" or in an "admin" directory?)

5. Also work for `from django.contrib.gis import admin` - ✅ #204 
   This imports the `register` decorator for convenience when authoring GIS apps.

adamchainz:
@UnknownPlatypus I have entertained myself tackling the two easiest ones here, perhaps you'd like to try the other two?

UnknownPlatypus:
@adamchainz Absolutely, would be nice to work on that and I think I have some ideas on how to tackle them.

adamchainz:
@UnknownPlatypus added point 5 above as well, for GIS admins.

adamchainz:
Alright we've checked everything off from this list! 🥳

This is now a really awesome feature that covers many cases. Great work @UnknownPlatypus .

UnknownPlatypus:
Thank again @adamchainz for the continuous support along the way and the kind help with my first few contributions to the open source community. :pray:

PR #200, opened 2022-09-14
Support multiple model classes in @admin.register fixer

UnknownPlatypus:
Addresses point number 2 (Multiple model registration) of #190 

Following a review on the initial @admin.register fixer, I've also updated documentation links to `/en/stable/releases/` to remove yellow/red banners.

### DIFF
--- a/src/django_upgrade/fixers/admin_register.py
+++ b/src/django_upgrade/fixers/admin_register.py
@@ -83,14 +83,14 @@ def update_class_def(
     tokens: list[Token], i: int, *, name: str, state: State, decorated: bool
 ) -> None:
     model_names = decorable_admins.get(state, {}).pop(name, set())
-    if len(model_names) == 1:
+    if model_names:
         if decorated:
             i = reverse_find(tokens, i, name=OP, src="@")
         j, indent = extract_indent(tokens, i)
         insert(
             tokens,
             j,
-            new_src=f"{indent}@admin.register({model_names.pop()})\n",
+            new_src=f"{indent}@admin.register({', '.join(sorted(model_names))})\n",
         )
 
 
@@ -130,17 +130,4 @@ def visit_Call(
         to_decorate = decorable_admins.get(state, {})
         if admin_name in to_decorate:
             to_decorate[admin_name].add(model_name)
-            yield ast_start_offset(node), partial(
-                erase_register_node,
-                node=parent,
-                admin_name=admin_name,
-                state=state,
-            )
-
-
-def erase_register_node(
-    tokens: list[Token], i: int, *, node: ast.Call, admin_name: str, state: State
-) -> None:
-    model_names = decorable_admins.get(state, {}).get(admin_name, set())
-    if len(model_names) == 1:
-        erase_node(tokens, i, node=node)
+            yield ast_start_offset(node), partial(erase_node, node=parent)
--- a/src/django_upgrade/fixers/admin_register.py
+++ b/src/django_upgrade/fixers/admin_register.py
@@ -125,7 +125,7 @@ def visit_Call(
 
         elif (  # admin.site.register((MyModel1, MyModel2), MyCustomAdmin)
             len(node.args) == 2
-            and isinstance((model_tuple := node.args[0]), ast.Tuple)
+            and isinstance((model_tuple := node.args[0]), (ast.Tuple, ast.List))
             and all(isinstance(elt, ast.Name) for elt in model_tuple.elts)
             and isinstance((admin_arg := node.args[1]), ast.Name)
             and not node.keywords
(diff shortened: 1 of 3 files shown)


src/django_upgrade/fixers/admin_register.py
```diff
@@ -108,39 +108,38 @@ def visit_Call(
...
+
+        elif (  # admin.site.register((MyModel1, MyModel2), MyCustomAdmin)
+            len(node.args) == 2
+            and isinstance((model_tuple := node.args[0]), ast.Tuple)
+            and all(isinstance(elt, ast.Name) for elt in model_tuple.elts)
+            and isinstance((admin_arg := node.args[1]), ast.Name)
+            and not node.keywords
+        ):
+            model_names = {
+                elt.id for elt in model_tuple.elts if isinstance(elt, ast.Name)
```
UnknownPlatypus, line 134: Mypy doesn't seems to narrow type correctly with `all(isinstance(elt, ast.Name) for elt in model_tuple.elts` so I had to do the explicit `isinstance(elt, ast.Name)` again. I don't know if you know any other way to workaround this ?

adamchainz:
> Following a review on the initial @admin.register fixer, I've also updated documentation links to `/en/stable/releases/` to remove yellow/red banners.

I'd prefer not to do this for documentation versions that are still online. It's better to link to the Django 3.1 version of the 3.1 release notes, rather than the Django 4.1 version of the 3.1 release notes, to avoid confusion. I only suggested it for Django 1.7 because it's offline and you found you needed to link to the `dev` version`, but `stable` is better there since it doesn't have the yellow "future version" banner. Can you undo the changes here? 😅

adamchainz:
Also in general please split PR's if they touch orthogonal areas of the code.

src/django_upgrade/fixers/admin_register.py
```diff
@@ -108,39 +108,38 @@ def visit_Call(
...
+            and isinstance((model_arg := node.args[0]), ast.Name)
+            and len(node.keywords) == 1
+            and node.keywords[0].arg == "admin_class"
+            and isinstance((admin_arg := node.keywords[0].value), ast.Name)
+        ):
+            model_names = {model_arg.id}
+
+        elif (  # admin.site.register((MyModel1, MyModel2), MyCustomAdmin)
+            len(node.args) == 2
+            and isinstance((model_tuple := node.args[0]), ast.Tuple)
```
adamchainz, line 128: We should also support lists

```suggestion
            and isinstance((model_tuple := node.args[0]), (ast.Tuple, ast.List))
```

tests/fixers/test_admin_register.py
```diff
@@ -470,6 +454,80 @@ class AuthorAdmin(CustomModelAdmin):
     )
 
 
+def test_multiple_model_multiline_registration():
```
adamchainz, line 457: let's add a duplicate of this test where the models are originally registered in the other order, to check the sorting behaviour

tests/fixers/test_admin_register.py
```diff
@@ -470,6 +454,80 @@ class AuthorAdmin(CustomModelAdmin):
...
+        @admin.register(MyModel1, MyModel2)
+        class MyCustomAdmin:
+            pass
+
+        """,
+        settings=settings,
+    )
+
+
+def test_multiple_model_tuple_registration():
```
adamchainz, line 482: add a duplicate of this test using a list of model classes

tests/fixers/test_admin_register.py
```diff
@@ -470,6 +454,80 @@ class AuthorAdmin(CustomModelAdmin):
...
+        from django.contrib import admin
+        from myapp.models import MyModel1, MyModel2
+
+        @admin.register(MyModel1, MyModel2, MyModel3)
+        class MyCustomAdmin:
+            pass
+
+        """,
+        settings=settings,
+    )
```
adamchainz, line 528: works nicely :)

adamchainz (commented):
Please update the docs example to show multiple models being combined into a single `admin.register()` call.

tests/fixers/test_admin_register.py
UnknownPlatypus: Yeah that's pretty cool :smiley:

src/django_upgrade/fixers/admin_register.py
```diff
@@ -83,14 +83,14 @@ def update_class_def(
...
-    if len(model_names) == 1:
+    if model_names:
         if decorated:
             i = reverse_find(tokens, i, name=OP, src="@")
         j, indent = extract_indent(tokens, i)
         insert(
             tokens,
             j,
-            new_src=f"{indent}@admin.register({model_names.pop()})\n",
+            new_src=f"{indent}@admin.register({', '.join(sorted(model_names))})\n",
```
adamchainz, line 93: imo squeezing too much into the f-string, will separate

PR #204, opened 2022-09-14
Support admin import from gis in @admin.register fixer

UnknownPlatypus:
Following https://github.com/adamchainz/django-upgrade/issues/190#issuecomment-1246776529.


adamchainz:
👍 👍 👍

Issue #224, opened 2022-09-16
`request.headers.get("Content-Length")` and `request.headers.get("Content-Type")`


christianbundy:
### Python Version

3.10

### Django Version

4.0

### Package Version

1.10

### Description

Hi! Thanks for your work on this project, I wish I'd found it sooner! I was just applying the fixers, and noticed that two HTTP headers aren't 'fixed' to user `request.headers.get()`:

- `CONTENT_LENGTH`
- `CONTENT_TYPE`

I was going to write a PR to fix this, but noticed that you have a test specifically ensuring that these are not fixed:

```python3
def test_not_header_access():
    check_noop(
        """\
        request.META['CONTENT_LENGTH']
        """,
        settings,
    )
```

Can you help me understand why these aren't fixed? I may be misunderstanding the docs:

> [`HttpRequest.headers`](https://docs.djangoproject.com/en/4.0/ref/request-response/#django.http.HttpRequest.headers)
> A case insensitive, dict-like object that provides access to all HTTP-prefixed headers (plus Content-Length and Content-Type) from the request.

Happy to help if you'd like these added to the fixer, but my gut feeling is that these were excluded intentionally and I'm misunderstanding something. Cheers.

adamchainz:
I think I just wrote that test as an example of something in `request.META` that should not be excluded. Seems like the fixer could certainly be expanded to support them, I'd just want a test updated to something else that's in `request.META` that shouldn't be rewritten (something from WSGI?).

Sounds like you've delved into the code a bit, want to write that PR? :)

christianbundy:
Thanks for the quick response! Yep, I'll open a PR here in a few minutes.

PR #226, opened 2022-09-16
Fix Content-Length and Content-Type

christianbundy:
Problem: Both Content-Length and Content-Type are HTTP headers, but they aren't picked up by the request header fixer because they don't start with the usual `HTTP_` prefix.

Solution: Add failing tests and then fix them by checking for either the usual `HTTP_` prefix or comparing against the set of these two exceptional cases.

Fixes: https://github.com/adamchainz/django-upgrade/issues/224

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,6 +14,10 @@ History
 
 * Add Django 2.0+ fixer to drop assignments of ``allow_tags`` attributes to ``True``.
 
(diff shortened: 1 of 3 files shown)


adamchainz:
Thank you! 👍 Needed a small bugfix with parentheses, you might want to check that out. I also added a changelog note - if you could try add that in your future open source contributions, that would save maintainers work.

christianbundy:
Ah, thanks for the paren bugfix -- I was just opening this PR to go fix that.

> [...] changelog note [...]

Will do!

PR #227, opened 2022-09-17
Improve timezone.utc fixer to reuse existing datetime imports

adamchainz:


ddaan:
This doesn't seem to work with imports that use parenthesis. This test currently fails.

```
def test_extend_datetime_multi_import():
    check_transformed(
        """\
        from datetime import (
            datetime, timedelta
        )
        from django.utils.timezone import utc
        do_a_thing(utc)
        """,
        """\
        from datetime import (
            datetime, timedelta, timezone
        )
        do_a_thing(timezone.utc)
        """,
        settings,
    )

```

```
>       assert fixed == dedented_after
E       AssertionError: assert 'from datetim...mezone.utc)\n' == 'from datetim...mezone.utc)\n'
E         Skipping 36 identical leading characters in diff, use -v to show
E         -  timedelta, timezone
E         - )
E         +  timedelta
E         + ), timezone
E           do_a_thing(timezone.utc)
```

ddaan:
How would we handle something like:
```
from datetime import timezone
print(timezone.now())
print(timezone.utc)
```

currently it rewritten to
```
from datetime import timezone
print(timezone.now())
print(timezone.utc)
```
what would break the timezone.now() implementation

adamchainz:
Thank you for your checks.

> This doesn't seem to work with imports that use parenthesis. This test currently fails.

Brilliant.

> How would we handle something like:

You're right, the current code assumes that `timezone` is used only for `utc`, which is flawed.  I think we can simply detect any use of `timezone.<not-utc` as a sign to leave the import.

PR #228, opened 2022-09-18
Support custom admin classes in @admin.register fixer

UnknownPlatypus:
For #190

Hey, so I started working on the custom admin support and I had a few interrogations to finish it. 

1. Is a stacked syntax ok when the same admin model is used to register a model against different sites ? 

``` diff
class MyAdminSite(admin.AdminSite):
    site_header = "My very custom administration"

custom_site = MyAdminSite(name="myadmin")

+@admin.register(Comment)
+@admin.register(Comment, site=custom_site)
class CommentAdmin(admin.ModelAdmin):
    pass


-admin.site.register(Comment, CommentAdmin)
-custom_site.register(Comment, CommentAdmin)
```

2. What should I do if `from django.contrib import admin` is missing (I suppose it's quitte common if you work with custom sites). I'm not sure where to add the import

3. Should we extend the `looks_like_admin_file` check heuristic to the regular fixers too ?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -151,6 +151,23 @@ __ https://docs.djangoproject.com/en/stable/ref/contrib/admin/#the-register-deco
     -admin.site.register(MyModel1, MyCustomAdmin)
     -admin.site.register(MyModel2, MyCustomAdmin)
 
+
+This also works with custom admin sites.
+We are detecting them heuristically by matching object names ending with ``site``, registered admin models ending with ``Admin`` and filenames called ``admin.py`` (or in an ``admin`` directory).
+
+.. code-block:: diff
+
+    from myapp.admin import custom_site
+    from django.contrib import admin
+
+    +@admin.register(MyModel)
+    +@admin.register(MyModel, site=custom_site)
+    class MyModelAdmin(admin.ModelAdmin):
+        pass
+
+    -custom_site.register(MyModel, MyModelAdmin)
+    -admin.site.register(MyModel, MyModelAdmin)
+
 Django 1.9
 -----------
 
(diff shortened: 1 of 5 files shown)


adamchainz:
> 1. Is a stacked syntax ok when the same admin model is used to register a model against different sites ?
> 
> 
> ```diff
> class MyAdminSite(admin.AdminSite):
>     site_header = "My very custom administration"
> 
> custom_site = MyAdminSite(name="myadmin")
> 
> +@admin.register(Comment)
> +@admin.register(Comment, site=custom_site)
> class CommentAdmin(admin.ModelAdmin):
>     pass
> -admin.site.register(Comment, CommentAdmin)
> -custom_site.register(Comment, CommentAdmin)
> ```

This seems fine to me - the decorator returns the admin class, unchanged, so that should work.

>     2. What should I do if `from django.contrib import admin` is missing (I suppose it's quitte common if you work with custom sites). I'm not sure where to add the import

I don't think we need to handle this case. It's idiomatic to import `admin` in order to subclass `admin.ModelAdmin`.

>     3. Should we extend the `looks_like_admin_file` check heuristic to the regular fixers too ?

I think it's fine to rely on the `admin` import in those cases.

UnknownPlatypus:
Great, thanks for the quick response.

Then I suppose you can check the code. 
Your answers are matching the assumptions I made :smile: 

If was thinking the changelog and docs are still relevant but let me know if you want me to add a mention somewhere of the different extensions to the initial fixer.

Cheers.

adamchainz:
I don't think the changelog needs updating but the readme could do with another example for cusotm sites, and maybe a sentence describing how custom sites are detected.

adamchainz:
Alright, looks good to merge. I've pushed a few small fixes, including support files with names like `custom_admin.py`. This is the most common pattern I've seen, where a projcet might have a main admin with classes in `admin.py` files, then a second admin site with its classes in `<name>_admin.py` files.

PR #229, opened 2022-09-18
Document ``path`` converter regexes and limitation

UnknownPlatypus:
I noticed while testing on a personal project that the url `slug` rewriter was matching the `[-a-zA-Z0-9_]+` but not the equivalent `[\w-]` or `[-\w]`.

I thought maybe we could support that too ?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -241,11 +241,6 @@ Rewrites imports of ``include()`` and ``url()`` from ``django.conf.urls`` to ``d
 .. |new path() syntax| replace:: new ``path()`` syntax
 __ https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax
 
-For some cases, this change alters the type of the arguments passed to the view, from ``str`` to the converted type (e.g. ``int``).
-This is not guaranteed backwards compatible: there is a chance that the view expects a string, rather than the converted type.
-But, pragmatically, it seems 99.9% of views do not require strings, and instead work with either strings or the converted type.
-Thus, you should test affected paths after this fixer makes any changes.
-
 .. code-block:: diff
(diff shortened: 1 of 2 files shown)


adamchainz:
Django’s `SlugConverter` uses that regex:

https://github.com/django/django/blob/f88fc72da4eb76f2d464edb4874ef6046f8a8658/django/urls/converters.py#L35-L36

`\w` matches more characters, since it matches all Unicode letter chars:

```
In [1]: import re

In [2]: re.match(r'\w', 'α')
Out[2]: <re.Match object; span=(0, 1), match='α'>
```

So unfortunately we can't automatically convert here. But I understand this might be quite common. Perhaps we could add some flag or extensibility here to allow saying that "in this project `\w` really only means to match latin letters" ?

UnknownPlatypus:
Oh indeed, the greek letters. 

Maybe a flag is a great idea if this case seems to be quite common ?
I'm not sure it actually is thought. Maybe we should keep the current behaviour after-all ?

adamchainz:
I think at least we could document the regexs, and maybe add a sentence that \w is sometimes used for slugs but it's not converted due to potential incompatibility. Would you like to take a crack at that?

UnknownPlatypus:
Looks like a good idea, I'll redo this PR tomorrow to document the slug regexes and why we don't support `\w`

UnknownPlatypus:
Took me more time than expected to get back to this, I'm in the middle of moving to a new place.
Anyway, let me know if it's looking good for you!

adamchainz:
Looking good, I just made a few tweaks to improve readability.

First commit merged here at the Djangocon Europe sprints 👍👍👍

PR #233, opened 2022-09-21
Rewrite admin.site.register() calls using model sequence and admin kwarg

adamchainz:
A little extra for #190.


adamchainz:
FYI @UnknownPlatypus 😊

PR #236, opened 2022-09-22
Extend `on_delete` fixer to support import from `django.db.models`

### LINKED ISSUE #62: Make on_delete fixer also work when fields imported *from* django.db.models
As noted in #61, we might want to also do e.g.

```diff
 from django.db.models import ForeignKey
+ from django.db.models import CASCADE

-ForeignKey()
+ForeignKey(on_delete=CASCADE)
```

UnknownPlatypus:
Fixes #62 

I also tweaked the `add_on_delete_keyword` function that was producing some extra spaces in some cases.
I've added tests to cover these cases.

### DIFF
--- a/src/django_upgrade/fixers/on_delete.py
+++ b/src/django_upgrade/fixers/on_delete.py
@@ -6,18 +6,65 @@ from __future__ import annotations
 
 import ast
 from functools import partial
-from typing import Iterable
+from typing import Iterable, MutableMapping
+from weakref import WeakKeyDictionary
 
 from tokenize_rt import Offset, Token
 
-from django_upgrade.ast import ast_start_offset
+from django_upgrade.ast import ast_start_offset, is_rewritable_import_from
 from django_upgrade.data import Fixer, State, TokenFunc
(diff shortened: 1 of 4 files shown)


src/django_upgrade/fixers/on_delete.py
```diff
@@ -6,61 +6,125 @@
...
+        used_names.add("CASCADE")
+
+        j, indent = extract_indent(tokens, i)
+        erase_node(tokens, i, node=node)
+        joined_names = ", ".join(sorted(used_names))
+        insert(
+            tokens,
+            j,
+            new_src=f"{indent}from django.db.models import {joined_names}\n",
+        )
```
adamchainz, line 67: A noble idea to try and nicely sort the imports, but unfortunately this will erase comments, and will undo the "one import per line" style

Generally users should be able to rely on their imoprt sorter (isort, reorder_python_imports) to fix up any non-compliant import style we add. So I'd rather do the most conservative definitely-will-work thing.

I am changing this to instead just insert `from django.db.models import CASCADE` on a line above.

We could also look at refactoring `update_imports` to add extra imports at the start, but that could be a separate issue.

Issue #238, opened 2022-09-24
Require at least one filename


adamchainz:
### Description

Currently running without any arguments does nothing, with no output:

```
$ django-upgrade
```

Kinda useless.

The filename arg should be updated to require at least one filename.

adamchainz:
@ddaan

adamchainz:
Fixed in #242.

PR #242, opened 2022-09-24
Require at least one filename

ddaan:
Fixes issue https://github.com/adamchainz/django-upgrade/issues/238

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -24,7 +24,7 @@ History
 
 * Expand the range of files considered settings files.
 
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -24,6 +24,10 @@ History
 
 * Expand the range of files considered settings files.
 
+* Require at least one filename
```
adamchainz, line 27:
```suggestion
* Require at least one filename.
```

PR #249, opened 2022-09-24
Make admin.register fixer skip unregistered models

### LINKED ISSUE #239: Avoid rewriting `admin.register()` to decorator after `unregister`
### Description

Currently the new `admin.register` fixer from #182 moves `register()` calls after `unregister()` calls:

```diff
+@admin.register(User)
 class CustomUserAdmin(...):
     ...

 admin.site.unregister(User)
-admin.site.register(User, CustomUserAdmin)
```

This is unsafe!

Update the fixer to detect `unregister` calls and avoid rewriting.

@andreasschmitz

andreasschmitz:
Updates the "register admin" fixer to ignore register calls if preceded by an unregister.

Closes #239

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,7 +197,6 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
 
     from django.contrib import admin
 
-
     class MyCustomAdmin(admin.ModelAdmin):
         ...
 
--- a/README.rst
+++ b/README.rst
@@ -197,8 +197,10 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
 
     from django.contrib import admin
 
+
     class MyCustomAdmin(admin.ModelAdmin):
         ...
+
 
     admin.site.unregister(MyModel1)
     admin.site.register(MyModel1, MyCustomAdmin)
--- a/README.rst
+++ b/README.rst
@@ -201,6 +201,7 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
     class MyCustomAdmin(admin.ModelAdmin):
         ...
 
+
     admin.site.unregister(MyModel1)
     admin.site.register(MyModel1, MyCustomAdmin)
 
--- a/README.rst
+++ b/README.rst
@@ -204,7 +204,6 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
     admin.site.unregister(MyModel1)
     admin.site.register(MyModel1, MyCustomAdmin)
 
-
 Django 1.9
 -----------
 
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,7 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
 
     from django.contrib import admin
 
+
     class MyCustomAdmin(admin.ModelAdmin):
         ...
 
--- a/README.rst
+++ b/README.rst
@@ -197,14 +197,11 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
 
     from django.contrib import admin
 
-
     class MyCustomAdmin(admin.ModelAdmin):
         ...
 
-
     admin.site.unregister(MyModel1)
     admin.site.register(MyModel1, MyCustomAdmin)
-
 
 
 Django 1.9
--- a/README.rst
+++ b/README.rst
@@ -201,7 +201,6 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
     class MyCustomAdmin(admin.ModelAdmin):
         ...
 
-
     admin.site.unregister(MyModel1)
     admin.site.register(MyModel1, MyCustomAdmin)
 
--- a/README.rst
+++ b/README.rst
@@ -197,10 +197,8 @@ If a ``register()`` call is preceded by an ``unregister()`` call that includes t
 
     from django.contrib import admin
 
-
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -171,6 +171,22 @@ Such calls are detected heuristically based on three criteria:
...
+
+    class MyCustomAdmin(admin.ModelAdmin):
+        ...
+
+
+    admin.site.unregister(MyModel1)
+    admin.site.register(MyModel1, MyCustomAdmin)
+
+
+
```
adamchainz, line 189:
```suggestion
```

README.rst
```diff
@@ -171,6 +171,22 @@ Such calls are detected heuristically based on three criteria:
     -custom_site.register(MyModel, MyModelAdmin)
     -admin.site.register(MyModel, MyModelAdmin)
 
+If a ``register()`` call is preceded by an ``unregister()`` call that includes the same model, it is ignored.
+
+.. code-block:: python
+
+    from django.contrib import admin
+
```
adamchainz, line 179:
```suggestion
```

README.rst
```diff
@@ -171,6 +171,22 @@ Such calls are detected heuristically based on three criteria:
...
+
+    class MyCustomAdmin(admin.ModelAdmin):
+        ...
+
+
```
adamchainz, line 184:
```suggestion
```

README.rst
aschmitz-wbd: Done. :)

README.rst
```diff
@@ -171,6 +171,20 @@ Such calls are detected heuristically based on three criteria:
...
+    from django.contrib import admin
+
+
+    class MyCustomAdmin(admin.ModelAdmin):
+        ...
+
+    admin.site.unregister(MyModel1)
+    admin.site.register(MyModel1, MyCustomAdmin)
+
+
```
adamchainz, line 187:
```suggestion
```

andreasschmitz:
Just realised that the pre-commit hooks unrolled the changes. Sorry for the confusion. How do you want to deal with it? 

Anything else I can do to move this along? The diff of the fixer isn't that nice, because everything is done in one function. 
Open for suggestions. Thx.

src/django_upgrade/fixers/admin_register.py
```diff
@@ -130,56 +133,103 @@ def visit_Call(
...
-
-        if (
-            admin_details is not None
-            and admin_details.parent == parents[-2]
-            and not (site_name and not admin_name.endswith("Admin"))
-        ):
-            admin_details.model_names_per_site.setdefault(site_name, set()).update(
-                model_names
+            and (  # unregister(…) has only one parameter
+                isinstance((first_arg := node.args[0]), ast.Name)
```
adamchainz, line 214: this assumes that the function is called with pos args, not kwargs

A line like:

```py
admin.site.unregister(model_or_iterable=MyModel1)
```

crashes this code, because there's no preceding check that `node.args` has a length of at least 1

I'm gonna update this code, whilst also doing the below update.

src/django_upgrade/fixers/admin_register.py
```diff
@@ -130,56 +133,103 @@ def visit_Call(
...
+                or (
+                    isinstance(first_arg, (ast.Tuple, ast.List))
+                    and all(isinstance(elt, ast.Name) for elt in first_arg.elts)
+                )
             )
```
adamchainz, line 219: The two args cases here are good for detecting when certain models are unregistered. But they don't handle the cases where the model names are generated dynamically, for example:

```python
admin.site.unregister(*some_models())
admin.site.register([MyModel1, MyModel2], MyCustomAdmin)
```

in such cases, I think we should consider *all* models unregistered, and thus not rewrite any following `register` calls. I'm gonna update the code to use a special representation for “maybe all models were unregistered”, and detect this.

adamchainz (commented):
Great, thank you for this. I didn't get around to looking at this in the sprints, this was a significant contribution that needed a bit more time to review than the other PR's. But conversely it's very valuable 👍

Gonnna add a couple fixes as per below.

adamchainz:
> Just realised that the pre-commit hooks unrolled the changes. Sorry for the confusion. How do you want to deal with it?
> 
> Anything else I can do to move this along? The diff of the fixer isn't that nice, because everything is done in one function. Open for suggestions. Thx.

I am happy with however the pre-commit hooks leave it. I just want minimum vertical whitespace to keep examples short.

adamchainz:
@andreasschmitz Thank you for your contribution 🙏 You're now one of “the most productive Django developers, maybe”.

@UnknownPlatypus Would you like to post-merge review this change? Maybe you can spot an issue.

UnknownPlatypus:
@adamchainz Looking good and the new tests seem to cover many cases so it should be nice. 
Thanks for the fix @andreasschmitz !

src/django_upgrade/fixers/admin_register.py
andreasschmitz: Oh, thank you. I've completely forgot about keyword arguments in this case.

src/django_upgrade/fixers/admin_register.py
andreasschmitz: That's a really good point. Thx a lot!

andreasschmitz:
Thx @adamchainz and @UnknownPlatypus

PR #251, opened 2022-09-25
Improve installation instructions

benjaoming:
I was reading the installation / usage and initially got a bit confused on which approach to choose and why.

This PR adds:

1. A more clear distinction between CLI and pre-commit usage
2. I also added some tips to clarify that it's not a pre-commit hook that one is supposed to simply remove again after the upgrade process
3. ~pre-commit configuration now says which file types to operate on (however, maybe that could/should be added in this repo's hook specification instead?)~ this is already defined in the hook definition, so no need!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,6 @@ You can install django-upgrade as a `pre-commit <https://pre-commit.com/>`__ hoo
         hooks:
         -   id: django-upgrade
             args: [--target-version, "4.1"]   # Replace with Django version
-            types: [file, python]
 
 To upgrade your entire project immediately, you can tell pre-commit to run the ``django-upgrade`` hook on all files. Once you commit these changes, your other pre-commit linters will execute and you may see further changes. Use the following command:
 
--- a/README.rst
+++ b/README.rst
@@ -23,13 +23,7 @@ Automatically upgrade your Django project code.
 Installation
 ============
 
-There are several ways to install and use django-upgrade, here are two common scenarios.
-
-
-Option 1: pip
--------------
-
-Install the ``django-upgrade`` command with pip. See **Usage** for an introduction to the ``django-upgrade`` command.
+Use **pip**:
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -40,14 +40,19 @@ Or with `pre-commit <https://pre-commit.com/>`__ in the ``repos`` section of you
         hooks:
         -   id: django-upgrade
             args: [--target-version, "4.1"]   # Replace with Django version
+            types: [file, python]
```
adamchainz, line 43: this is already in the hook spec: https://github.com/adamchainz/django-upgrade/blob/main/.pre-commit-hooks.yaml#L6

And I want to keep it there - since we'll add support for template files at some point..

Also `file` is also not needed - it's not possible for a 'python directory' in identify - `python` is based on extension, which is only detected for files:

https://github.com/pre-commit/identify/blob/8a6ab8dd14c708770a999b76d2b929af06266c1b/identify/identify.py#L40-L64

README.rst
```diff
@@ -40,14 +40,19 @@ Or with `pre-commit <https://pre-commit.com/>`__ in the ``repos`` section of you
...
 
-
-To upgrade your entire project run:
+To upgrade your entire project immediately, you can tell pre-commit to run the ``django-upgrade`` hook on all files. Once you commit these changes, your other pre-commit linters will execute and you may see further changes. Use the following command:
 
 .. code-block:: sh
 
     pre-commit run django-upgrade --all-files
 
+.. tip:: Leaving this hook in place after the upgrade process is considered entirely safe and actually recommended! Django-upgrade may receive future improvements and thus keep improving your code.
```
adamchainz, line 51: the tip block doesn't have any special rendering on GitHub so I'd rather just use a paragraph which starts with bold “tip”

README.rst
```diff
@@ -40,14 +40,19 @@ Or with `pre-commit <https://pre-commit.com/>`__ in the ``repos`` section of you
...
+
+.. tip:: Normally, pre-commit will run exactly on the files that you have changed in your commit. Pre-commit can also help you update to the latest ``django-upgrade`` version with its ``autoupdate`` command. This is a sufficient flow for most cases, but remember to also run ``pre-commit run django-upgrade --all-files`` as django-upgrade may add new improvements in each release.
```
adamchainz, line 53: I think we don't need to explain a lot about how pre-commit works here, just a sentence at most.

Also it's always lower case "pre-commit"

README.rst
```diff
@@ -23,15 +23,15 @@ Automatically upgrade your Django project code.
 Installation
 ============
 
-Use **pip**:
+**Option 1**: Install django-upgrade with **pip**:
```
adamchainz, line 26: maybe we could use real headings rather than “option 1” / ”option 2”?

README.rst
benjaoming: Great, will remove! I somehow missed the line :facepalm:

README.rst
benjaoming: Agreed, didn't have time to review the GitHub rendering, will fix this.

README.rst
benjaoming: Sure, I couldn't figure out what to write between the "Installation" heading and "Option 1...". I prefer at paragraph between to headings. Will add some simple one-liner.

adamchainz:
I reformatted it a bit more and reduced it down to avoid repeating what’s said in the pre-commit docs.

PR #255, opened 2022-09-26
Stop django.urls fixer rewriting imports in a loop

### LINKED ISSUE #250: Undesired rewriting of import statements (pre-commit loop)
### Python Version

3.8

### Django Version

2.2

### Package Version

main

### Description

I need some help triaging this behavior, as I'm not familiar with the patterns that django-upgrade uses to detect and alter import statements.

While running django-upgrade on a project, I have the following import statement generated:

```python
from django.urls import include, path, re_path
```

However, if my project is using `reorder_python_imports`, it will split up this line into 3 import statements:

```python
from django.urls import include
from django.urls import path
from django.urls import re_path
```

Now, if I run the `django-upgrade` hook again, the former line reappears like this:

```python
from django.urls import include
from django.urls import path
from django.urls import include, path, re_path
```

You can add the latest version of `reorder_python_imports` with this configuration:

```
-   repo: https://github.com/asottile/reorder_python_imports
    rev: v3.8.3
    hooks:
    -   id: reorder-python-imports
        language_version: python3

```

Are you able to immediately grasp the issue @adamchainz or does it need to be isolated into a test project / test case?

### The correct behavior?

I think that what I would expect from django-upgrade is that it generates whatever import statements that it needs, but that subsequent linting from `black`, `flake8` or `reorder_python_imports` is left in place and doesn't result in any new changes.

adamchainz:
Fixes #250.

### DIFF
--- a/src/django_upgrade/fixers/django_urls.py
+++ b/src/django_upgrade/fixers/django_urls.py
@@ -52,11 +52,12 @@ def visit_ImportFrom(
         and is_rewritable_import_from(node)
         and any(alias.name == "re_path" for alias in node.names)
     ):
-        yield ast_start_offset(node), partial(
-            update_django_urls_import,
-            node=node,
-            state=state,
-        )
+        pass
+        # yield ast_start_offset(node), partial(
+        #     maybe_remove_re_path,
+        #     node=node,
+        #     state=state,
+        # )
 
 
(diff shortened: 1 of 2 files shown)


adamchainz:
Fixed in #270

PR #258, opened 2022-09-28
Add fixer for `invalid_list` message of `ModelMultipleChoiceField`

### LINKED ISSUE #256: Replace `ModelMultipleChoiceField`'s `list` error message with `invalid_list`
### Description

Since django 3.1, the [forms.ModelMultipleChoiceField](https://docs.djangoproject.com/en/3.1/ref/forms/fields/#modelmultiplechoicefield) `list` error message key is [deprecated](https://docs.djangoproject.com/en/3.1/releases/3.1/#id2) in favor of `invalid_list`.

SImilarly to the `EmailValidator(whitelist=...)`, I think this could be auto-fixed.

```diff
-forms.ModelMultipleChoiceField(error_messages={"list": "Enter multiple values."})
+forms.ModelMultipleChoiceField(error_messages={"invalid_list": "Enter multiple values."})
```
Would you be interested in a PR ?

UnknownPlatypus:
Fixes #256

### DIFF
--- a/src/django_upgrade/fixers/forms_model_multiple_choice_field.py
+++ b/src/django_upgrade/fixers/forms_model_multiple_choice_field.py
@@ -1,0 +1,53 @@
+"""
+Replace `list` error message key with `list_invalid` for ModelMultipleChoiceField.
+https://docs.djangoproject.com/en/3.1/releases/3.1/#id2
+"""
+from __future__ import annotations
+
+import ast
+from functools import partial
+from typing import Iterable
+
+from tokenize_rt import Offset
+
+from django_upgrade.ast import ast_start_offset
+from django_upgrade.data import Fixer, State, TokenFunc
+from django_upgrade.tokens import replace
+
+fixer = Fixer(
+    __name__,
+    min_version=(3, 1),
+)
+
+
+@fixer.register(ast.Call)
+def visit_Call(
+    state: State,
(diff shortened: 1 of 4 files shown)


src/django_upgrade/fixers/forms_model_multiple_choice_field.py
```diff
@@ -0,0 +1,53 @@
...
+            )
+        )
+        and any(
+            (error_message_node := kw).arg == "error_messages" for kw in node.keywords
+        )
+        and isinstance(error_message_node.value, ast.Dict)
+        and any(
+            (isinstance(key, ast.Constant) and (list_node := key).value == "list")
+            for key in error_message_node.value.keys
+        )
```
adamchainz, line 51: I think this is the first time I've seen walrus used inside a comprehension!

adamchainz (approved):

adamchainz:
Brilliant, thank you for yet another star contribution.

src/django_upgrade/fixers/forms_model_multiple_choice_field.py
UnknownPlatypus: It's actually a neet trick to find a culprit using the `any` short circuit behaviour. Not so many use cases thought but it works nicely in this case.

PR #259, opened 2022-09-28
Remove domain_allowlist fix for EmailValidator

UnknownPlatypus:
While working on the `ModelMultipleChoiceField` fixer, I noticed that the `EmailValidator` was allowed to rewrite `domain_whitelist` kwarg to `domain_allowlist` however this is not a valid kwarg for an `EmailValidator`.

This is only an undocumented attribute to the `EmailValidator` class. From the docs:

> The whitelist parameter is deprecated. Use [allowlist](https://docs.djangoproject.com/en/4.0/ref/validators/#django.core.validators.EmailValidator.allowlist) instead. The undocumented domain_whitelist attribute is deprecated. Use domain_allowlist instead.

This PR simply remove the `domain_whitelist` fix and update the docs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -685,7 +685,7 @@ This only applies in command files, which are heuristically detected as files wi
 ``EmailValidator``
 ~~~~~~~~~~~~~~~~~~
 
-Rewrites keyword arguments to their new names: ``whitelist`` to ``allowlist``, and ``domain_whitelist`` to ``domain_allowlist``.
+Rewrites keyword arguments to their new names: ``whitelist`` to ``allowlist``.
 
 .. code-block:: diff
 
@@ -693,8 +693,6 @@ Rewrites keyword arguments to their new names: ``whitelist`` to ``allowlist``, a
 
     -EmailValidator(whitelist=["example.com"])
     +EmailValidator(allowlist=["example.com"])
-    -EmailValidator(domain_whitelist=["example.org"])
-    +EmailValidator(domain_allowlist=["example.org"])
 
 ``default_app_config``
 ~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -685,16 +685,14 @@ This only applies in command files, which are heuristically detected as files wi
 ``EmailValidator``
 ~~~~~~~~~~~~~~~~~~
 
-Rewrites keyword arguments to their new names: ``whitelist`` to ``allowlist``, and ``domain_whitelist`` to ``domain_allowlist``.
+Rewrites keyword arguments to their new names: ``whitelist`` to ``allowlist``.
```
adamchainz, line 688:
```suggestion
Rewrites the ``whitelist`` keyword argument to its new name ``allowlist``.
```

src/django_upgrade/tokens.py
```diff
@@ -368,7 +368,6 @@ def replace_argument_names(
 
     for n, keyword in reversed(list(enumerate(node.keywords))):
         if keyword.arg in arg_map:
-            x, y = kwarg_func_args[n]
```
adamchainz, line 371: ah looks like flake8 doesn't detect unused vars after a multiple assignment...

adamchainz (approved):
Ah woops, thanks!

I guess we could rewrite the attribute in subclasses, but that seems rather niche and probably not worth doing.

README.rst
UnknownPlatypus: You didn't applied this suggestion before merging, not sure it was intended ?

adamchainz: woops!

adamchainz: done in 76819494042fc3aaaa01604ed40418f4ce0078b0

PR #260, opened 2022-09-30
Preserve double quotes when rewriting string tokens

kevinmarsh:
I was attempting to use `django-upgrade` on a project which prefers double quotes over single quotes, meaning when something was rewritten like:
```diff
-request.META["HTTP_SERVER"]
+request.headers['Server']
```
it's linter would complain that it's expecting double quotes.

I'm not sure this solution is fully comprehensive but wanted to get feedback before spending any more time on it, but hopefully it's not too controversial to keep the existing quote style during rewrites.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -42,6 +42,10 @@ History
 
   Thanks to Benjamin Bach for the report in `Issue #250 <https://github.com/adamchainz/django-upgrade/issues/250>`__, and to Thibaut Decombe for the fix in `PR #270 <https://github.com/adamchainz/django-upgrade/pull/270>`__.
 
+* Fixers that modify string literals now match existing use of double quotes.
+
+  Thanks to Kevin Marsh in `PR #260 <https://github.com/adamchainz/django-upgrade/pull/260>`__.
+
 * Make fixers that erase lines also erase any trailing comments.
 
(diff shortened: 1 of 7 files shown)


adamchainz:
Thanks for the report, but I do not want to pursue this route. As the readme says:

> django-upgrade focuses on upgrading your code and not on making it look nice. Run django-upgrade before formatters like [Black](https://black.readthedocs.io/en/stable/).

Pretty sure 99% of users will just run Black afterwards. Myabe you do too if you want double quoted strings. The pre-commit install instructions also tell you how to place django-upgrade to do this.

adamchainz:
Please in future open an issue for discussion before doing any development.

Btw there is a good reason for using repr(), it handles adding escaping of special characters automatically. There's no double quoted version of it, so at best we could convert the repr, but it feels error prone and kinda pointless when code format gets already exist.

kevinmarsh:
No worries, I figured that'd be the case of not wanting to inspect the user's code more than you have to. 

> before doing any development

I created the PRs rather than issues since I still wanted to still use `django-upgrade` on that particular project to get it updated (done with a forked version) and just to show what the diff could look like, so my time wasn't wasted and hopefully yours wasn't too much...

adamchainz:
On second thought I think this is a good idea.

PR #261, opened 2022-09-30
Preserve tab indentation when rewriting decorators

kevinmarsh:
Similar to #260, if a project also prefers tabs over spaces, currently `django-upgrade` will transform

into this containing tabs and spaces:

This is simple enough to resolve when the line being re-writtnen already contains tabs (see passing test) but would require a larger refactor for lines that aren't indented (see `xfail`-ing test). Am leaving as a draft but could be merged as a partial fix, or just closed as a "don't want to go down this rabbit hole"

### DIFF
--- a/src/django_upgrade/fixers/admin_decorators.py
+++ b/src/django_upgrade/fixers/admin_decorators.py
@@ -168,6 +168,7 @@ def decorate_function(
     if decorated:
         i = reverse_find(tokens, i, name=OP, src="@")
     j, indent = extract_indent(tokens, i)
+    default_indent = indent if indent else "    "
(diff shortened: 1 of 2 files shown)


adamchainz:
Do you use tabs or are you just pursuing perfection? Again like #260 django-upgrade doesn't focus on formatting just outputting working code.

Python doesn't mind mixed tab/space indentation [except where it would be ambiguous](https://docs.python.org/3/reference/lexical_analysis.html#indentation):

> Indentation is rejected as inconsistent if a source file mixes tabs and spaces in a way that makes the meaning dependent on the worth of a tab in spaces; a [TabError](https://docs.python.org/3/library/exceptions.html#TabError) is raised in that case.

I don't think this tabs + 4 spaces example could cause that.

PR #270, opened 2022-10-18
Fix urls fixer wrong import rewrite

### LINKED ISSUE #250: Undesired rewriting of import statements (pre-commit loop)

UnknownPlatypus:
My attempt to fix #250.

I noticed that the underlying issue is that some noop tests are actually doing rewrites.

For exemple, the following is actually triggering an erase/rewrite of the import node but since it rewrites the same thing we don't notice it:

```python
def test_fake_noop():
    check_noop(
        """\
        from django.urls import re_path

        re_path('whatever')
        """,
        settings,
    )
```

I added a `state_added_names` variable to keep track of names that will need to be imported. Then I can only add an import with these names (minus the ones initially imported). 

PS: Last commit replace the `state_used_names` variable with `state_re_path_used` because it was only needed to determine if the `re_path` import can be removed.

I'm unsure if it's the best way to deal with this so I'm happy to take any feed back. At least maybe this will give you some food for thoughts on your attempt.

Cheers

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -36,6 +36,10 @@ History
 
 * Update ``timezone.utc`` fixer to only use absolute references from existing imports of the ``datetime`` module.
 
+* Make Django 2.0+ URL fixer avoid a loop of adding imports that already exist.
+
+  Thanks to Benjamin Bach for the report in `Issue #250 <https://github.com/adamchainz/django-upgrade/issues/250>`__, and to Thibaut Decombe for the fix in `PR #270 <https://github.com/adamchainz/django-upgrade/pull/270>`__.
+
 * Make fixers that erase lines also erase any trailing comments.
 
 * Fix leaving a trailing comma when editing imports in certain cases.
--- a/src/django_upgrade/fixers/django_urls.py
+++ b/src/django_upgrade/fixers/django_urls.py
@@ -17,7 +17,6 @@ from django_upgrade.compat import str_removeprefix
 from django_upgrade.data import Fixer, State, TokenFunc
 from django_upgrade.tokens import (
     STRING,
-    erase_node,
     extract_indent,
     find,
     insert,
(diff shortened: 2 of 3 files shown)


adamchainz:
Thank you. I will review this properly in the next few days, then hopefully do a django-upgrade release with all the built-up stuff!

> PS: Last commit replace the `state_used_names` variable with `state_re_path_used` because it was only needed to determine if the `re_path` import can be removed.
> 
> I'm unsure if it's the best way to deal with this so I'm happy to take any feed back. At least maybe this will give you some food for thoughts on your attempt.

This seems like a good approach to me!

UnknownPlatypus:
Woops, forgot about the release note, my bad. Anyway, thanks for the review and the few tweaks !

adamchainz:
Thank you! Release coming shortly, writing a blog post for it...

UnknownPlatypus:
Awesome! I saw that you have been very productive lately with the blog posts. Really nice! Keep up the good work 😃

sshishov:
Thank you @adamchainz and @UnknownPlatypus for fixing it in a short period of time 🎉

PR #279, opened 2022-10-27
Make `re_path` fixer convert more `include()`s

### LINKED ISSUE #263: Make `re_path` -> `path` fixer convert `include()`s with unterminated regexes
### Python Version

3.10.7

### Django Version

4.1.2

### Package Version

1.10.0

### Description

Not sure if this is a bug strictly speaking or a feature request, but after running this package with `--target-version 4.1` the fixer that converts `re_path` usage to `path` misses cases which should result in a direct translation.

An example from our codebase is
`re_path(r"^accounts/", include("allauth.urls")),`

In the current version this line is left unmodified.

Presumably something like this could be automatically converted to
`path("accounts/", include("allauth.urls")),`

UnknownPlatypus:
Fixes #263.

Now that the blocking issue is resolved, here is a patch! Let me know if you need me to make some adjustments.

_PS: I needed the `removesuffix` utility so I took the one from the [PEP 616](https://peps.python.org/pep-0616/#specification) similarly to what you probably did for the `removeprefix` one._

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,8 +2,12 @@
 History
 =======
 
+* Make ``re_path`` -> ``path`` fixer also convert ``include()``'s with unterminated regexes.
+
+  Thanks to Thibaut Decombe in `PR #279 <https://github.com/adamchainz/django-upgrade/pull/279>`__.
+
 * Avoid rewriting ``request.META`` to ``request.headers`` in ``del`` statements.
-  This pattern works for ``request.META`` but not for ``request.headers`` which is an immutable Mapping.
+  This pattern works for ``request.META`` but not for ``request.headers`` which is an immutable mapping.
 
   Thanks to Thibaut Decombe in `PR #290 <https://github.com/adamchainz/django-upgrade/pull/290>`__.
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,7 +2,7 @@
 History
 =======
 
-* Make ``re_path`` -> ``path`` fixer also convert ``include()``'s with unterminated regexes.
+* Make ``re_path`` -> ``path`` fixer also convert ``include()``\'s with unterminated regexes.
 
   Thanks to Thibaut Decombe in `PR #279 <https://github.com/adamchainz/django-upgrade/pull/279>`__.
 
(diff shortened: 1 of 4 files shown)


src/django_upgrade/fixers/django_urls.py
```diff
@@ -154,9 +154,20 @@ def visit_Call(
             ):
                 regex_path = node.args[0].value
 
+            include_called = (
+                len(node.args) > 1
```
adamchainz, line 158:
```suggestion
                len(node.args) >= 2
```

adamchainz (commented):
Could you rebase now #280 is merged?

> PS: I needed the removesuffix utility so I took the one from the [PEP 616](https://peps.python.org/pep-0616/#specification) similarly to what you probably did for the removeprefix one.

Yup that's what I did 👍

Everything here looks logical but will double review tomorrow...

UnknownPlatypus:
> Everything here looks logical but will double review tomorrow...

ofc, no rush. Thanks for the quick review :raised_hands: 

I've juste rebased on main.

HISTORY.rst
```diff
@@ -2,6 +2,10 @@
 History
 =======
 
+* Make ``re_path`` -> ``path`` fixer also convert ``include()``'s with unterminated regexes.
```
adamchainz, line 5: reStructuredText requires an escape if putting something right next to inline syntax

looks like GitHub’s renderer is fine with it but PyPI’s won't be
```suggestion
* Make ``re_path`` -> ``path`` fixer also convert ``include()``\'s with unterminated regexes.
```

src/django_upgrade/fixers/django_urls.py
```diff
@@ -195,10 +213,11 @@ def fix_url_call(
 }
 
 
-def convert_path_syntax(regex_path: str) -> str | None:
-    if not regex_path.endswith("$"):
+def convert_path_syntax(regex_path: str, include_called: bool) -> str | None:
+    if not regex_path.endswith("$") and not include_called:
```
adamchainz, line 217: Apply [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to make this clearer

```suggestion
    if not (regex_path.endswith("$") or include_called):
```

tests/fixers/test_django_urls.py
```diff
@@ -213,6 +213,39 @@ def test_re_path_unanchored_end():
     )
 
 
+def test_re_path_unanchored_end_with_include():
```
adamchainz, line 216: This end is anchored
```suggestion
def test_re_path_anchored_end_with_include():
```

adamchainz (approved):

src/django_upgrade/fixers/django_urls.py
UnknownPlatypus: Probably also slightly faster with the `or` short circuit possibility

PR #280, opened 2022-10-27
Update database engine postgresql

### LINKED ISSUE #268: Update settings.DATABASES backend path `django.db.backends.postgresql_psycopg2`
This backend was renamed in Django 1.9. There’s still a compat path:
https://github.com/django/django/blob/84206607d6bfd61e7f7a88b51163ffd4153e3b5a/django/db/utils.py#L108-L110

But better to update outdated settings so they’re in line with docs and actual module layout.

UnknownPlatypus:
Another small one :smile:.

Fixes #268


src/django_upgrade/fixers/settings_database_posgresql.py
```diff
@@ -0,0 +1,50 @@
...
+            (
+                isinstance(key, ast.Constant)
+                and key.value == "ENGINE"
+                and isinstance(val, ast.Constant)
+                and (target_node := val).value
+                == "django.db.backends.postgresql_psycopg2"
+            )
+            for key, val in zip(node.keys, node.values)
+        )
+        and state.looks_like_settings_file()
```
adamchainz, line 46: we can go a bit faster by putting this settings file check first (and caching it: #281 ), since it's the most likely check to exclude a node from consideration

src/django_upgrade/fixers/settings_database_posgresql.py
```diff
@@ -0,0 +1,50 @@
...
+    ):
+        yield ast_start_offset(target_node), partial(
+            replace, src='"django.db.backends.postgresql"'
+        )
```
adamchainz, line 50: since #260, django-upgrade tries to match the quoting choice of existing code, so updating this to use a small token function to match the existing quote char

tests/fixers/test_settings_database_posgresql.py
```diff
@@ -0,0 +1,133 @@
...
+                "PASSWORD": "mypassword",
+                "HOST": "127.0.0.1",
+                "PORT": "5432",
+            }
+        }
+        """,
+        settings,
+        filename="myapp/settings.py",
+    )
+
```
adamchainz, line 97: We should also test when `DATABASES` has two relevant backends, I'm inserting a test here

src/django_upgrade/fixers/settings_database_posgresql.py
```diff
@@ -0,0 +1,52 @@
+"""
```
adamchainz: the name of the file was missing “t” 😉

src/django_upgrade/fixers/settings_database_posgresql.py
```diff
@@ -0,0 +1,52 @@
...
+    node: ast.Dict,
+    parents: list[ast.AST],
+) -> Iterable[tuple[Offset, TokenFunc]]:
+    if (
+        state.looks_like_settings_file()
+        and isinstance(parents[-1], ast.Dict)
+        and isinstance((db_setting := parents[-2]), ast.Assign)
+        and isinstance(db_setting.targets[0], ast.Name)
+        and db_setting.targets[0].id == "DATABASES"
+        and len(db_setting.targets) == 1
```
adamchainz: length checks should be done before accessing list indexes, to avoid crashes where the list isn't the required length. I don't think it's possible to have a crash here, though:

* `parents[-1]` will always exist, and if it's a `Dict`, `parents[-2]` will always exist, at least as the `ast.Module`
* I don't think it's possible for an `Assign` to have 0 targets (would be weird)

But still worth putting the length checks first to embed the right pattern, plus they should be cheaper than the `isinstance` checks.

adamchainz:
One small concern I just had - this will break code like:

```py
if settings.DATABASES["default"]["ENGINE"] == "django.db.backends.postgresql_psycopg2":
    ...
```

and there’s not really much we can do to discover that (could be looked at in arbitrary variables)

I *think* that’srather unlikely though, since the recommended way to check a DB vendor is:

```py
from django.db import connection
if connection.vendor == "postgres":
    ...
```

WDYT?

src/django_upgrade/fixers/settings_database_posgresql.py
UnknownPlatypus: > I don't think it's possible for an Assign to have 0 targets (would be weird)

That's the assumption I made and the reason why I delayed the `len` check but yeah I'm not sure it's faster anyway and it's less clear so let's keep your version.

> parents[-1] will always exist, and if it's a Dict, parents[-2] will always exist, at least as the ast.Module

yeah exactly, but I agree it might be confusing. Better explicit than implicit and the performance cost wont matter anyway

src/django_upgrade/fixers/settings_database_posgresql.py
UnknownPlatypus: Woops :sweat_smile:

UnknownPlatypus:
yeah it looks rather unlikely I agree.

Also the [first](https://stackoverflow.com/questions/18846174/django-detect-database-backend) [few](https://stackoverflow.com/questions/32815765/how-to-check-which-database-is-being-used-in-a-django-project) stack overflow question i found about that were generally recommandation the `connection.vendor` way.

Maybe you should ask the folks on twitter which one they would have used to get more feedback on that ? 

I feel like the recommended way is maybe not that well documented so I expect most people to land on stack overflow if they want to know how to check the project database vendor.

src/django_upgrade/fixers/settings_database_posgresql.py
UnknownPlatypus: ah yeah right, I've seen in the blog post you merge that in the end. mb

adamchainz:
> Also the [first](https://stackoverflow.com/questions/18846174/django-detect-database-backend) [few](https://stackoverflow.com/questions/32815765/how-to-check-which-database-is-being-used-in-a-django-project) stack overflow question i found about that were generally recommandation the `connection.vendor` way.

Good research.

> Maybe you should ask the folks on twitter which one they would have used to get more feedback on that ?

I think this is kinda niche, likely won't see many responses.

Anyway we trust users to still be adults. If they use django-upgrade to change this string they’re responsible for any downstream effects.

PR #286, opened 2022-11-03
Use cross-platform regexes for some `State.looks_like*` functions

### LINKED ISSUE #283: NullBooleanField upgraded to BooleanField in migrations file
### Python Version

3.10

### Django Version

3.2.5 --> 4.1

### Package Version

1.11.0

### Description

While using django upgrade on the entire django project using the following in the main django project folder (powershell):
`git ls-files -- '*.py' | %{django-upgrade --target-version 4.1 $_}`

Also the migration files where targeted (one of the files before conversion, located at projects > migrations):

```python:
# Generated by Django 2.1.7 on 2020-04-24 10:49

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        #('projects', '0237_auto_20200420_0955'),
        ('projects', '0002_auto_20200512_1030'),
    ]

    operations = [
        migrations.AddField(
            model_name='historicalsequenceequipment',
            name='cond_data',
            field=models.NullBooleanField(default=False),
        ),
        migrations.AddField(
            model_name='sequenceequipment',
            name='cond_data',
            field=models.NullBooleanField(default=False),
        ),
    ]

```
As the docs says that these were not included in the NullBooleanField change & the code checks for state.looks_like_migrations_file in this fixer, this seemed like a bug to me ("migrations" is in the file path as it is the folder name)

Thanks for the package, I'm loving it!
William

CLWilliam:
Fixes #283
- `looks_like_migrations_file` and `looks_like_command_file` are converted to regex based filename checks.
- Tests for  `looks_like_migrations_file` added
- git example added for powershell to apply django-upgrade on multiple files

### DIFF
--- a/src/django_upgrade/data.py
+++ b/src/django_upgrade/data.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import ast
-import os
 import pkgutil
 import re
 from collections import defaultdict
@@ -29,7 +28,9 @@ class Settings:
 
 
 admin_re = re.compile(r"(\b|_)admin(\b|_)")
+commands_re = re.compile(r"(^|[\\/])management[\\/]commands[\\/]")
 dunder_init_re = re.compile(r"(^|[\\/])__init__\.py$")
+migrations_re = re.compile(r"(^|[\\/])migrations([\\/])")
 settings_re = re.compile(r"(\b|_)settings(\b|_)")
 test_re = re.compile(r"(\b|_)tests?(\b|_)")
 
@@ -53,12 +54,7 @@ class State:
 
     @cached_property
     def looks_like_command_file(self) -> bool:
-        parts = self.filename.split(os.path.sep)
-        try:
-            i = parts.index("commands")
-        except ValueError:
-            return False
-        return i > 0 and i < (len(parts) - 1) and parts[i - 1] == "management"
+        return commands_re.search(self.filename) is not None
 
     @cached_property
     def looks_like_dunder_init_file(self) -> bool:
@@ -66,7 +62,7 @@ class State:
 
     @cached_property
     def looks_like_migrations_file(self) -> bool:
-        return "migrations" in self.filename.split(os.path.sep)
+        return migrations_re.search(self.filename) is not None
 
     @cached_property
     def looks_like_settings_file(self) -> bool:
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -15,6 +15,10 @@ History
 
   Thanks to Thibaut Decombe in `PR #280 <https://github.com/adamchainz/django-upgrade/pull/280>`__.
 
+* Make detection of management commands and migration files detect both forward and backward slashes as directory separators.
+
+  Thanks to William Claassen in `PR #286 <https://github.com/adamchainz/django-upgrade/pull/286>`__.
+
 1.11.0 (2022-10-26)
 -------------------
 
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -86,7 +86,8 @@ For more on usage run ``django-upgrade --help``.
 Run django-upgrade before formatters like `Black <https://black.readthedocs.io/en/stable/>`__.
 
 ``django-upgrade`` does not have any ability to recurse through directories.
-Use the pre-commit integration, globbing, or another technique for applying to many files such as |with git ls-files pipe xargs|_. For example:
+Use the pre-commit integration, globbing, or another technique for applying to many files such as |with git ls-files pipe xargs|_.
+For example when using ``bash``/``sh``:
```
adamchainz, line 90:
```suggestion
For example with a POSIX shell (``bash/``, ``zsh``, etc.):
```

README.rst
```diff
@@ -95,6 +96,14 @@ Use the pre-commit integration, globbing, or another technique for applying to m
 
     git ls-files -- '*.py' | xargs django-upgrade --target-version 4.1
 
+
+Or when using PowerShell:
```
adamchainz, line 100:
```suggestion
…or with PowerShell:
```

src/django_upgrade/data.py
```diff
@@ -33,6 +32,8 @@ def __init__(self, target_version: tuple[int, int]) -> None:
 dunder_init_re = re.compile(r"(^|[\\/])__init__\.py$")
 settings_re = re.compile(r"(\b|_)settings(\b|_)")
 test_re = re.compile(r"(\b|_)tests?(\b|_)")
+migrations_re = re.compile(r"(^|[\\/])migrations([\\/])")
+commands_re = re.compile(r"(^|[\\/])management[\\/]commands[\\/]")
```
adamchainz, line 36: please sort the regexes alphabetically, the same order that the methods are in

The commands method did deliberately check the current file lives within `commands`, which the regex currently doesn’t. I think we could modify it like so to capture this:

```
commands_re = re.compile(r"(^|[\\/])management[\\/]commands[\\/][^\\/]+$")
```

Since you’re adjusting `commands_re`, please ensure it is tested with both `\` and `/`

adamchainz (changes requested):
Thanks this is looking good.

Please add a changelog note in `HISTORY.rst` following the existing style (whenever you make an open source contribution.

src/django_upgrade/data.py
CLWilliam: This one fails on `myapp/subapp/management/commands/test/subcommand.py` as it has a subfolder and thus an additional path separator. expanding it to 
```
(^|[\\/])management[\\/]commands[\\/][^\\/]+[\\/]*[^\\/]*$
```
captures one subfolder. Pushing it further to 0 or more subfolders brings: 
```
(^|[\\/])(management[\\/]commands[\\/])([^\\/]+[\\/]+)*([^\\/]+)$
```
--> regexr.com/71kmn

Commands tests are also updated for windows paths.

adamchainz: Yeah, that was a deliberate suggestion - subdirectories won't be picked up as management commands by Django, so we can ignore certain fixers there.

adamchainz: Ah actually my bad, I didn't realize `subcommand` was in the existing tests. We can keep support for `subcommand`. Reverting to your original easier regex since the 0+ folders search is verbose and doesn’t help much.

adamchainz:
Thank you!

Issue #74, opened 2021-09-23
HttpRequest.headers fixer produces unsupported assigments in tests


browniebroke:
### Python Version

3.8

### Django Version

3.1

### Package Version

1.3.1

### Description

I ran django-upgrade on our codebase and it replaced some places where we were setting `request.META` to a fake requests created by [`RequestFactory`](https://docs.djangoproject.com/en/stable/topics/testing/advanced/#django.test.RequestFactory) in our tests, which isn't correct.

I guess `django-upgrade` shouldn't change these if they are on the left hand side of an assignment.

Here is an example:

```diff
  request = RequestFactory().get('/')
- request.META['HTTP_USER_AGENT'] = user_agent
+ request.headers['User-Agent'] = user_agent
```

When running tests, it causes: `TypeError: 'HttpHeaders' object does not support item assignment`. 

### known workarounds

1. Rename the variable to something else than `request`, e.g. `req`
2. In this specific example, I was able to avoid django-upgrade to touch our code if I rewrite it with:

   ```diff
   - request = RequestFactory().get('/')
   - request.META['HTTP_USER_AGENT'] = user_agent 
   + request = RequestFactory(HTTP_USER_AGENT=user_agent).get('/')
   ```

adamchainz:
Ah, I did not know that one could assign to request.META in tests

We can heuristically avoid touching test requests by checking if the node's parent is a function whose name starts with `test_`...

(Also assigning to `request.headers` in tests sounds like a useful feature for Django to add, if you want to find/make a ticket there.)

adamchainz:
> I guess `django-upgrade` shouldn't change these if they are on the left hand side of an assignment.

I missed this. This is easy to do.

adamchainz:
Fixed in #75, relaesed in version 1.3.2.

browniebroke:
Nice! 🚀

PR #290, opened 2022-11-08
Make request.headers avoid rewriting `del` statements

UnknownPlatypus:
Similar to what happened in #74, I ran into this issue:

```diff
-del request.META["HTTP_REFERER"]
+del request.headers["Referer"]
```

Which raises: `'HttpHeaders' object does not support item deletion`

I guess this is probably an odd pattern but we used it to add a temporary header in a middleware that we wanted to clean at the end. I added a small change to make sure we don't rewrite in that case.

Let me know if it feels right to you.
Cheers

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,10 @@
 =======
 History
 =======
+* Avoid rewriting ``request.META`` to ``request.headers`` in ``del`` statements.
+  This pattern works for ``request.META`` but not for ``request.headers`` which is an immutable Mapping.
+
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -1,6 +1,10 @@
 =======
 History
 =======
+* Avoid rewriting ``request.META`` to ``request.headers`` in ``del`` statements.
```
adamchainz, line 4:
```suggestion

* Avoid rewriting ``request.META`` to ``request.headers`` in ``del`` statements.
```

adamchainz (approved):
Great, thanks!

UnknownPlatypus:
Nice, thanks for the fast review again !

Issue #271, opened 2022-10-22
Django 4.2: Rewrite test request factory and client usage to use `headers`


David-Wobrock:
### Description

Following https://github.com/django/django/pull/16163, a new `headers` parameter was introduced in 4.2 for the 4 classes:
- `RequestFactory`
- `Client`
- `AsyncRequestFactory`
- `AsyncClient`

Let's use that parameter instead of the keyword arguments, which were prefixed with `HTTP_` for the synchronous classes (WSGI).

adamchainz:
Basically rewrites like this:

```diff
-self.client.get("/", HTTP_USER_AGENT="curl/123", SCRIPT_NAME="/app/")
+self.client.get("/", headers={"user-agent": "curl/123"), SCRIPT_NAME="/app/")
```

PR #305, opened 2022-12-22
Add HTTP Headers fixer for Client and RequestFactory.

David-Wobrock:
Fixes https://github.com/adamchainz/django-upgrade/issues/271

At the moment, it's just a first draft, to get the idea.
Some cases are not handled yet, and we will have to decide how far to push the handled cases :)

Used for https://github.com/django/django/pull/16401


src/django_upgrade/fixers/http_headers.py
```diff
@@ -0,0 +1,185 @@
...
+HEADERS_KWARG = "headers"
+HTTP_PREFIX = "HTTP_"
+
+
+def merge_http_headers_kwargs(tokens: list[Token], i: int, *, node: ast.Call) -> None:
+    http_kwargs_names = []
+    contains_headers_kwarg = False
+
+    for kw in node.keywords:
+        if kw.arg and kw.arg.startswith(HTTP_PREFIX):
```
David-Wobrock, line 38: What about `CONTENT_TYPE` and `CONTENT_LENGTH`?
Should they also be re-written to `headers` kwarg?

src/django_upgrade/fixers/http_headers.py
```diff
@@ -0,0 +1,185 @@
...
+
+@fixer.register(ast.Call)
+def visit_Call(
+    state: State,
+    node: ast.Call,
+    parents: list[ast.AST],
+) -> Iterable[tuple[Offset, TokenFunc]]:
+    if (
+        isinstance(node.func, ast.Name)
+        and "RequestFactory" in state.from_imports["django.test"]
```
David-Wobrock, line 165: `AsyncRequestFactory` and `AsyncClient` are not handled at the moment, as the keyword arguments don't start with `HTTP_` I guess they would be harder to catch 🤔

tests/fixers/test_http_headers.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_transform_constructor_call_multiple():
+    check_transformed(
+        """\
+        from django.test import Client
+        c = Client("/", HTTP_HOST="test.com", headers={"user-agent": "example"}, HTTP_ACCEPT_LANGUAGE="fr-fr")
+        """,  # noqa: E501
+        """\
+        from django.test import Client
+        c = Client("/", headers={"user-agent": "example", "host": "test.com", "accept-language": "fr-fr"},)
```
David-Wobrock, line 32: Trailing comma here, is that an issue? 🤔 

Or we remove it when the `headers` is the last argument.

tests/fixers/test_http_headers.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_transform_constructor_call():
+    check_transformed(
+        """\
+        from django.test import RequestFactory
+        RequestFactory("/", HTTP_HOST="test.com")
+        """,
+        """\
+        from django.test import RequestFactory
+        RequestFactory("/", headers={"host": "test.com"})
```
David-Wobrock, line 18: This works, but it might be harder to catch cases like
```
RequestFactory().get("/", HTTP_HOST="host.com")
```
Should we handle them?

tests/fixers/test_http_headers.py
```diff
@@ -0,0 +1,105 @@
...
+            "/password_reset/",
+            {"email": "staffmember@example.com"},
+            HTTP_HOST="www.example:dr.frankenstein@evil.tld",
+        )
+        """,
+        """
+        response = self.client.post(
+            "/password_reset/",
+            {"email": "staffmember@example.com"},
+            headers={"host": "www.example:dr.frankenstein@evil.tld"}
```
David-Wobrock, line 101: This doesn't work yet 😁

src/django_upgrade/fixers/http_headers.py
```diff
@@ -0,0 +1,205 @@
...
+from typing import Iterable
+
+from tokenize_rt import Offset
+from tokenize_rt import Token
+
+from django_upgrade.ast import ast_start_offset
+from django_upgrade.data import Fixer
+from django_upgrade.data import State
+from django_upgrade.data import TokenFunc
+from django_upgrade.fixers.assert_form_error import looks_like_client_call
```
David-Wobrock, line 18: Should we move this to a common module?
What would be a good place? :)

src/django_upgrade/fixers/test_http_headers.py
```diff
@@ -0,0 +1,189 @@
...
+            Token(name="OP", src=","),
+            Token(name="UNIMPORTANT_WS", src=" "),
+        ]
+        new_tokens += [
+            Token(name="STRING", src=f'"{key}"'),
+            Token(name="OP", src=":"),
+            Token(name="UNIMPORTANT_WS", src=" "),
+        ]
+        new_tokens += value
+    tokens[headers_closing_idx:headers_closing_idx] = new_tokens
```
adamchainz, line 146: when injecting new code we don't need to build up loads of individual tokens, we can just create a single token with the target source.

src/django_upgrade/fixers/test_http_headers.py
```diff
@@ -0,0 +1,189 @@
...
+    headers_closing_idx = find(tokens, headers_arg_start_idx, name=OP, src="}")
+
+    new_tokens = []
+    for key, value in kwarg_dict.items():
+        new_tokens += [
+            Token(name="OP", src=","),
+            Token(name="UNIMPORTANT_WS", src=" "),
+        ]
+        new_tokens += [
+            Token(name="STRING", src=f'"{key}"'),
```
adamchainz, line 141: use `str_repr_matching` in case of necessary escaping

src/django_upgrade/fixers/test_http_headers.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+        if ends_with_whitespace:
+            arg_end_idx += 1
+
+        del tokens[kwarg_name_start_idx:arg_end_idx]
+        return erased_tokens, kwarg_name_start_idx
+    return [], -1
+
+
+def insert_into_existing_headers_kwarg(
```
adamchainz, line 126: I'm a little apprehensive about adding this pathway. Just a lot more to test. I suppose it can "fix up" attempts to use the old way on upgraded code...

src/django_upgrade/fixers/test_http_headers.py
```diff
@@ -0,0 +1,189 @@
...
+def visit_Call(
+    state: State,
+    node: ast.Call,
+    parents: list[ast.AST],
+) -> Iterable[tuple[Offset, TokenFunc]]:
+    if (
+        isinstance(node.func, ast.Name)
+        and node.func.id in ("Client", "RequestFactory")
+        and node.func.id in state.from_imports["django.test"]
+    ) or (looks_like_client_call(node, "client") and node.args):
```
adamchainz, line 45: filter for keyword argument here, to avoid queueing a token rewriting functions for every client call, on every run

adamchainz (commented):
Just been looking into this now I have the time, and Django 4.2 alpha 1 is out. I've made some fixes, comments on outstanding changes to make.

tests/fixers/test_http_headers.py
adamchainz: I don't think so, we can't reliably detect common patterns of using `RequestFactory`

src/django_upgrade/fixers/http_headers.py
adamchainz: Yes

adamchainz: Actually maybe not? https://github.com/django/django/blob/69352d85fa8412865db9e0c7f177b333c0eac3e2/django/test/client.py#L577-L582

adamchainz:
Okay I’ve rewritten the implementation to resolve all noted issues.

Tested against Django before your fixup commit, Sentry, and some of my clients' projects. All worked well on those.

David-Wobrock:
Thanks a lot for your help @adamchainz 🙇 
Happy to see this merged 🙌

PR #313, opened 2023-01-17
Rename `lookup_needs_distinct` to `lookup_spawns_duplicates`

browniebroke:
Function is undocumented and was therefore renamed without deprecation in Django 4.0:

https://docs.djangoproject.com/en/4.0/releases/4.0/#miscellaneous

> The undocumented `django.contrib.admin.utils.lookup_needs_distinct()` function is renamed to `lookup_spawns_duplicates()`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


adamchainz:
Thanks! Another small fix towards greatness.

PR #333, opened 2023-03-06
Add option to select specific fixer to run

### LINKED ISSUE #278: Option to select specific fixer to run
### Description

It would be nice to have the option to specify one or more fixer to apply. 
One usecase is to be able to categorise the changes in several different commits, and another usecase is the codebase might not be ready to apply some of the fixes but others.

davidszotten:
Add `--fixer foo` option (can be repeated) to only run specific fixers. Fixes #278


src/django_upgrade/data.py
```diff
@@ -178,9 +181,14 @@ def _import_fixers() -> None:
 _import_fixers()
 
 
-def get_ast_funcs(target_version: tuple[int, int]) -> ASTCallbackMapping:
+def get_ast_funcs(
+    target_version: tuple[int, int], fixers: list[str] | None
+) -> ASTCallbackMapping:
     ast_funcs: ASTCallbackMapping = defaultdict(list)
     for fixer in FIXERS:
+        # TODO: is this where we want to do the name extraction?
```
adamchainz, line 189: no

tests/test_main.py
```diff
@@ -104,6 +104,33 @@ def test_main_stdin_no_changes(capsys):
...
+
+    out, err = capsys.readouterr()
+    assert err == ""
+
+    if changed:
+        assert result == 1
+        assert out == "from django.core.paginator import Paginator\n"
+    else:
+        assert result == 0
+        assert out == input_
```
adamchainz, line 131: I don't like conditional tests, it makes the test code too complicated. can we instead have two test functions, one, for the "changes expected case", one for the "no changes expected"?

adamchainz (changes requested):
please add changelog note, and update the readme to include fixer names

i'll review this properly later

src/django_upgrade/main.py
```diff
@@ -44,6 +44,9 @@ def main(argv: Sequence[str] | None = None) -> int:
             "4.2",
         ],
     )
+    parser.add_argument(
+        "--fixers", action="append", help="only apply the specified fixer(s)"
```
ulgens, line 48: I think this would be a bit easier to read with a comma at the end.

adamchainz (changes requested):
Thank you for starting this @davidszotten. I deferred reviewing because I want this to come in not as a piecemeal feature, but fully formed so it covers most use cases.

I have just rebased the branch and plan to continue work later. I think we should still add the following:

- [ ] Rename the option to `--only`, since it selects *only* some fixers (or some other suggestion?)
- [ ] Add an opposing `--skip` option to skip named fixers. This would users work around when one fixer makes bad changes in their code base.
- [ ] Add a `--list-fixers` option to list the fixers.
- [ ] Documentation for all the above.

gav-fyi:
Hi, all. 

I was just wondering what the current state of this PR is?

I can't tell if it's currently with David, Adam, or looking for volunteers to implement the additions mentioned above (happy to volunteer if it's the latter).

This feature would have come in useful recently as I upgraded a few projects to 5.0.

adamchainz:
I'm not actively working on it and it doesn't seem like David is either. If you want to make a new PR extending this one with the remaining pieces that I listed, I'd be glad of the help.

gav-fyi:
Sorry for the spam ☝🏻 I forgot I'd referenced this PR early on in my WIP PR description. 😬

adamchainz:
Done in #443.

PR #338, opened 2023-03-17
Fix admin.register custom site ordering bug

### LINKED ISSUE #337: admin_site.register -> @admin.register results in "undefined name 'admin_site'"
### Python Version

3.8.13

### Django Version

4.1.7

### Package Version

1.12.0

### Description

Django-upgrade correctly transforms the `admin_site.register(StrandingSpeciesPhoto, StrandingSpeciesPhotoAdmin)` into the decorator-form`@admin.register(StrandingSpeciesPhoto, site=admin_site)`, but `admin_site` is defined after the decorator, resulting in `NameError: name 'admin_site' is not defined`.

This could be fixed by moving the definition of `admin_site` up, but I'm not sure how feasible that is. 

```diff
+ @admin.register(StrandingSpeciesPhoto, site=admin_site)
  class StrandingSpeciesPhotoAdmin(admin.ModelAdmin):
      list_display = ["__str__", "thumbnail"]
      autocomplete_fields = ["species"]
@@ -20,5 +21,4 @@ def thumbnail(self, obj):


  admin_site = StrandingAdminSite(name="stranding")
- admin_site.register(StrandingSpeciesPhoto, StrandingSpeciesPhotoAdmin)
  admin_site.register(StrandingSpecies, SpeciesAdmin)
```

adamchainz:
Fixes #337.

### DIFF
--- a/src/django_upgrade/fixers/admin_register.py
+++ b/src/django_upgrade/fixers/admin_register.py
@@ -36,11 +36,11 @@ fixer = Fixer(
 
 
 class AdminDetails:
-    __slots__ = ("parent", "node", "model_names_per_site")
+    __slots__ = ("parent", "lineno", "model_names_per_site")
 
-    def __init__(self, parent: ast.AST, node: ast.ClassDef) -> None:
+    def __init__(self, parent: ast.AST, lineno: int) -> None:
         self.parent = parent
-        self.node = node
+        self.lineno = lineno
         self.model_names_per_site: dict[str, set[str]] = {}
 
 
@@ -67,7 +67,7 @@ def visit_ClassDef(
 ) -> Iterable[tuple[Offset, TokenFunc]]:
     if _is_django_admin_imported(state) and not uses_full_super_in_init_or_new(node):
         decorable_admins.setdefault(state, {})[node.name] = AdminDetails(
-            parents[-1], node
+            parents[-1], node.lineno
         )
         if not node.decorator_list:
             offset = ast_start_offset(node)
@@ -213,7 +213,7 @@ def visit_Call(
                             )
                         )
                         is not None
-                        and site_defined_line < admin_details.node.lineno
+                        and site_defined_line < admin_details.lineno
                     )
                 )
             ):
--- a/src/django_upgrade/fixers/admin_register.py
+++ b/src/django_upgrade/fixers/admin_register.py
@@ -36,10 +36,11 @@ fixer = Fixer(
 
 
 class AdminDetails:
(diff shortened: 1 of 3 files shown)


adamchainz:
Currently a PoC. This approach works but is quite a lot of stuff.

Another direction we could pursue is only rewriting if the custom site is imported at the top of the module. That would be easier although it would miss some cases.

If we do adopt the approach in this PR, it would be best to make the "defined names" thing general, so we can reuse it elsewhere. Perhaps the top-level `visit()` function could track defined names, rather than `from_imports` - this would add some flexibilty to fixers.

@UnknownPlatypus do you have any input?

src/django_upgrade/fixers/admin_register.py
```diff
@@ -36,10 +36,11 @@
 
 
 class AdminDetails:
-    __slots__ = ("parent", "model_names_per_site")
+    __slots__ = ("parent", "node", "model_names_per_site")
 
-    def __init__(self, parent: ast.AST) -> None:
+    def __init__(self, parent: ast.AST, node: ast.ClassDef) -> None:
         self.parent = parent
+        self.node = node
```
UnknownPlatypus, line 43: Could maybe only track `node.lineno` ? It seems like the only thing you actually use later?

UnknownPlatypus:
> Another direction we could pursue is only rewriting if the custom site is imported at the top of the module. That would be easier although it would miss some cases.

Ye, I think tracking line number for this case might be okay, this is quite an usual case with line of codes being moved quite a lot.

> If we do adopt the approach in this PR, it would be best to make the "defined names" thing general, so we can reuse it elsewhere. Perhaps the top-level visit() function could track defined names, rather than from_imports - this would add some flexibilty to fixers.

This might add quite an overhead for a feature we rarely use (might be needed for the `MyChoices.choices` -> `MyChoices` rewrite tho). But if you see good use case / simplifications that could rely on that, it might be a good idea. Will need to handle correctly scopes and redefined names

Otherwise, maybe just a `get_FOO_defined_line` function we can sparingly use for those special cases is enough (basically what you did with `get_site_defined_line` ?

adamchainz:
> This might add quite an overhead for a feature we rarely use (might be needed for the `MyChoices.choices` -> `MyChoices` rewrite tho). But if you see good use case / simplifications that could rely on that, it might be a good idea. Will need to handle correctly scopes and redefined names

Yeah I was thinking we'd just track module-level names, to reduce overhead of scoping and chance for bugs. Better to be a bit conservative.

PR #369, opened 2023-07-11
Drop `.choices` from model field choices

### LINKED ISSUE #336: Django 5.0: Drop `.choices` for model field `choices` parameters
### Description

https://code.djangoproject.com/ticket/34388

```diff
 class Card(models.Model):
-    suit = models.IntegerField(choices=Suit.choices)
+    suit = models.IntegerField(choices=Suit)
```

UnknownPlatypus:
Fixes #336 

This might break if someone designed a custom class with a `choices` property which is not a django enumeration class but I don't think this is really a use case ?
The documentation links are not yet valid, I think we have to play the waiting game when this MR seems ready for you.

Cheers

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-* Support Django 5.0 as a target version. No fixers yet.
+* Support Django 5.0 as a target version.
 
 * Add Django 5.0+ fixer to drop ``.choices`` on model field ``choices`` parameters when using an enumeration type.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,10 @@ Changelog
 
 * Support Django 5.0 as a target version. No fixers yet.
 
+* Add Django 5.0+ fixer to drop ``.choices`` on model field ``choices`` parameters when using an enumeration type.
+
+  Thanks to Thibaut Decombe in `PR #369 <https://github.com/adamchainz/django-upgrade/pull/369>`__.
+
 1.14.1 (2023-08-16)
 -------------------
 
--- a/README.rst
+++ b/README.rst
@@ -910,4 +910,13 @@ Django 5.0
 
 `Release Notes <https://docs.djangoproject.com/en/5.0/releases/5.0/>`__
 
-No fixers yet.
+``.choices``
+~~~~~~~~~~~~
+
+Drop ``.choices`` for model field ``choices`` parameters. This parameter now accept enumeration types directly.
+
+.. code-block:: diff
+
+     class Card(models.Model):
+    -    suit = models.IntegerField(choices=Suit.choices, default=Suit.DEFAULT)
+    +    suit = models.IntegerField(choices=Suit, default=Suit.DEFAULT)
--- a/README.rst
+++ b/README.rst
@@ -910,10 +910,11 @@ Django 5.0
 
 `Release Notes <https://docs.djangoproject.com/en/5.0/releases/5.0/>`__
 
-``.choices``
-~~~~~~~~~~~~
+Model field ``choices=<name>.choices``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Drop ``.choices`` for model field ``choices`` parameters. This parameter now accept enumeration types directly.
+Drop ``.choices`` for model field ``choices`` parameters.
+This parameter now accept enumeration types directly.
 
 .. code-block:: diff
 
--- a/README.rst
+++ b/README.rst
@@ -914,7 +914,7 @@ Model field ``choices=<name>.choices``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Drop ``.choices`` for model field ``choices`` parameters.
-This parameter now accept enumeration types directly.
+This parameter now accepts enumeration types directly.
 
 .. code-block:: diff
 
--- a/src/django_upgrade/fixers/model_field_choices.py
+++ b/src/django_upgrade/fixers/model_field_choices.py
@@ -16,6 +16,8 @@ from django_upgrade.data import Fixer
 from django_upgrade.data import State
 from django_upgrade.data import TokenFunc
 from django_upgrade.tokens import find_last_token
+from django_upgrade.tokens import OP
+from django_upgrade.tokens import reverse_find
 
 fixer = Fixer(
     __name__,
@@ -51,4 +53,5 @@ def visit_Call(
 
 def remove_choices(tokens: list[Token], i: int, node: ast.Attribute) -> None:
     j = find_last_token(tokens, i, node=node)
-    del tokens[i + 1 : j + 1]
+    i = reverse_find(tokens, j, name=OP, src=".")
+    del tokens[i : j + 1]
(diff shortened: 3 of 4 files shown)


UnknownPlatypus:
> perhaps copy-pasting in the old versions of the Choices classes as a backport, then never upgrading.

Indeed this is more likely. Maybe it's time to have some sort of configuration added to select rules ? I'm thinking of 2 possible ways of doing it:
- Adding a way to select/exclude some fixers when running django-upgrade. This would also cover https://github.com/adamchainz/django-upgrade/pull/333. We could also disable some maybe unsafe rules by default and require user opt-in.
- Adding some sort of applicability flag (`--unsafe`/`--safe`), we could then flag individual fixers as `safe` or not and only apply safe fixers by default. One could enable everything using the `--unsafe` flag. (I'm not sure I really like this option, I feel like It lacks some granularity compared to the other option)

> it’s probably okay, but this is also a very tiny improvement, especially since the choices attribute is being retained without deprecation.

I agree this is a tiny improvement but since it has been added to django core, I think it's good to push this feature (and I do think it's more user-friendly from an API perspective.

tests/fixers/test_model_field_choices.py
```diff
@@ -0,0 +1,123 @@
...
+        """,
+        settings,
+    )
+
+
+def test_untransformed_in_migration_file():
+    check_noop(
+        """\
+        from django.db.models import CharField
+        CharField(choices=[(1, 2), (3, 4)]) # No .choices in migrations anyway
```
adamchainz, line 24: move the comment out of the test file

src/django_upgrade/fixers/model_field_choices.py
```diff
@@ -0,0 +1,58 @@
...
+                isinstance(node.func, ast.Attribute)
+                and "models" in state.from_imports["django.db"]
+                and isinstance(node.func.value, ast.Name)
+                and node.func.value.id == "models"
+            )
+            or (
+                isinstance(node.func, ast.Name)
+                and node.func.id in state.from_imports["django.db.models"]
+            )
+        )
```
adamchainz, line 45: I think it's actually very limiting to only act on fields from `django.db.models`. Custom fields exist in many places, like `django.contrib.postgres`, third party apps, and even projects. Let's actually limit this to calls to things with a name ending in `Field` ? Need to handle names and chains of `Attribute`s like `django.db.models.CharField(...)`.

src/django_upgrade/fixers/model_field_choices.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@fixer.register(ast.Call)
+def visit_Call(
+    state: State,
+    node: ast.Call,
+    parents: list[ast.AST],
+) -> Iterable[tuple[Offset, TokenFunc]]:
+    if (
+        not state.looks_like_migrations_file
```
adamchainz, line 33: Let's instead introduce `looks_like_models_file` which checks for the name `models.py` or a directory `models` in the path. That'll be more precise and avoid running this on many unnecessary function calls.

adamchainz (commented):
Thanks for rebasing after I added the 5.0 support.

Returning to the conversation, I think this would actually be pretty safe. Looking again, I think you were probably right that the chances are slim that anyone actually has a custom class with `.choices`. Let’s go ahead as is and fix if anyone reports issues.

I have some comments about expanding the scope, if you wouldn't mind.

README.rst
```diff
@@ -910,4 +910,13 @@ Django 5.0
 
 `Release Notes <https://docs.djangoproject.com/en/5.0/releases/5.0/>`__
 
-No fixers yet.
+``.choices``
+~~~~~~~~~~~~
```
adamchainz, line 914: Slightly broader name

```suggestion
Model field ``choices=<name>.choices``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

UnknownPlatypus:
Thanks the thorough review, will look into it later this week :ok_hand:

src/django_upgrade/fixers/model_field_choices.py
UnknownPlatypus: Should I do the same in `null_boolean_field.py` ?

UnknownPlatypus: I went ahead and did that, let me know if you think I should revert that

tests/test_data.py
```diff
@@ -179,6 +179,45 @@ def test_looks_like_migrations_file_false(filename: str) -> None:
     assert not state.looks_like_migrations_file
 
 
+@pytest.mark.parametrize(
+    "filename",
+    (
+        "my_app/models/blog.py",
+        "my_app\\models\\blog.py",
```
adamchainz, line 186: Let's test the two-deep case as well
```suggestion
        "my_app/models/blog.py",
        "my_app\\models\\blog.py",
        "my_app/models/blogging/blog.py",
        "my_app\\models\\blogging\\blog.py",
```

src/django_upgrade/fixers/null_boolean_field.py
```diff
@@ -36,7 +36,7 @@ def visit_ImportFrom(
     parents: list[ast.AST],
 ) -> Iterable[tuple[Offset, TokenFunc]]:
     if (
-        not state.looks_like_migrations_file
+        state.looks_like_models_file
```
adamchainz, line 39: Good work. I think I'll actually split `looks_like_models_file` and this change into a separate PR.

CHANGELOG.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 
 * Support Django 5.0 as a target version. No fixers yet.
```
adamchainz, line 5:
```suggestion
* Support Django 5.0 as a target version.
```

README.rst
```diff
@@ -910,4 +910,14 @@ Django 5.0
 
 `Release Notes <https://docs.djangoproject.com/en/5.0/releases/5.0/>`__
 
-No fixers yet.
+Model field ``choices=<name>.choices``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Drop ``.choices`` for model field ``choices`` parameters.
+This parameter now accept enumeration types directly.
```
adamchainz, line 917:
```suggestion
This parameter now accepts enumeration types directly.
```

tests/fixers/test_model_field_choices.py
```diff
@@ -0,0 +1,169 @@
...
+        """,
+        """\
+        from django.contrib.postgres.fields import ArrayField
+        field = ArrayField(choices=Card)
+        """,
+        settings,
+        filename="models/blog.py",
+    )
+
+
```
adamchainz, line 77: unnecessary after change to `ending "Field"
```suggestion
```

tests/fixers/test_model_field_choices.py
```diff
@@ -0,0 +1,169 @@
...
+def test_transform_with_kwarg_ending_comma():
```
adamchainz, line 78: wanna add a test where the enum type is referred to like `enums.Card.choices`

adamchainz (commented):
Great work, gonna do final edits and merge

tests/fixers/test_model_field_choices.py
adamchainz: good thing I did, it fixed a bug where too much was erased :)

PR #423, opened 2024-02-06
Add Django 1.10+ request.user attribute fixer

### LINKED ISSUE #419: Convert is_authenticated() to is_authenticated
### Description

Having used `django-upgrade` to bring a few applications up from `1.11`, it's been great but the upgrade scripts skipped rewriting `user.is_authenticated()` as a bool `user.is_authenticated`. Since the change was made final in Django, it should probably rewrite those references as well.

Alurith:
Fixes #419


src/django_upgrade/fixers/request_user_is_authenticated .py
```diff
@@ -0,0 +1,67 @@
...
+                and node.value.attr == "request"
+            )
+        )
+    )
+
+
+def rewrite_user_is_auth(tokens: list[Token], i: int) -> None:
+    j = find(tokens, i, name=NAME, src="is_authenticated")
+    y = find(tokens, i, name=OP, src=")")
+    del tokens[j + 1 : y + 1]
```
adamchainz, line 67: Find the opening parenthesis and delete from the opening to closing only, rather than `j+1`. There may be space, even newlines and comments, before the parenthesis!

Add these tests to see what I mean:

```py
request . user . is_authenticated ( )
```


```py
if (
    request
    .user
    .is_authenticated  # bla
    ()
):
    ...
```

Alurith: Thanks for the feedback, I also found a whitespace in the fixer's filename that I removed.

I am working on a fix for the additional cases you suggested but I encountered an issue.
I've changed the rewrite function like this:
```python
def rewrite_user_is_auth(tokens: list[Token], i: int) -> None:
    j = find(tokens, i, name=NAME, src="is_authenticated")
    y = find(tokens, j, name=OP, src="(")
    z = find(tokens, y, name=OP, src=")")
    del tokens[z]
    del tokens[y]
```
but I can't make it work with this case
```python
def test_spaces_and_comments():
    check_transformed(
        """\
        if (
            request
            .user
            .is_authenticated  # bla
            ()
        ):
            ...
        """,
        """\
        if (
            request
            .user
            .is_authenticated  # bla
            
        ):
            ...
        """,
        settings,
    )

```
This is the output of pytest
```shell
>       assert fixed == dedented_after
E       AssertionError: assert 'if (\n    request\n    .user\n    .is_authenticated  # bla\n    \n):\n    ...\n' == 'if (\n    request\n    .user\n    .is_authenticated  # bla\n    ()\n):\n    ...\n'
E
E           if (
E               request
E               .user
E               .is_authenticated  # bla
E         -     ()
E         ?     --
E         +     
E           ):
E               ...

tests\fixers\tools.py:25: AssertionError
```
It seems the only different between the two string are some whitespaces after the comment, so I tought it was because of my formatter but even disabling it or thinkering with the whitespaces didn't solve the issue.

I am sure I am missing something here, can you point me in the right direction?

adamchainz: The problem is gonna be because either Black or `textwrap.dedent()` trims trailing whitespace in the file, whilst the fixer leaves it. We’re fine leaving the whitespace, django-upgrade leaves formatting to Black or whatever.

This version of the test should work:

```py
def test_spaces_and_comments():
    check_transformed(
        """\
        if (
            request
            .user
            .is_authenticated  # bla
            () # bla
        ):
            ...
        """,
        """\
        if (
            request
            .user
            .is_authenticated  # bla
             # bla
        ):
            ...
        """,
        settings,
    )
```

README.rst
```diff
@@ -321,6 +321,15 @@ Whilst mentioned in the `Django 2.1 release notes <https://docs.djangoproject.co
     -from django.contrib.staticfiles.templatetags.staticfiles import static
     +from django.templatetags.static import static
 
+``user.is_authenticated()``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Rewrites ``user.is_authenticated()`` in ``user.is_authenticated``.
```
adamchainz, line 327:
```suggestion
Rewrites calls to ``request.user.is_authenticated()`` and ``request.user.is_anonymous()`` to remove the parentheses, per `the deprecation <https://docs.djangoproject.com/en/1.10/releases/1.10/#using-user-is-authenticated-and-user-is-anonymous-as-methods>`__.
```

README.rst
```diff
@@ -321,6 +321,15 @@ Whilst mentioned in the `Django 2.1 release notes <https://docs.djangoproject.co
     -from django.contrib.staticfiles.templatetags.staticfiles import static
     +from django.templatetags.static import static
 
+``user.is_authenticated()``
```
adamchainz, line 324:
```suggestion
``request.user`` boolean attributes
```

README.rst
```diff
@@ -321,6 +321,15 @@ Whilst mentioned in the `Django 2.1 release notes <https://docs.djangoproject.co
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Rewrites ``user.is_authenticated()`` in ``user.is_authenticated``.
+
+.. code-block:: diff
+
+    -user.is_authenticated()
+    +user.is_authenticated
```
adamchainz, line 332: This diff is inaccurate - `request.user` or `self.request.user` are required. Let’s show one of each with both:

```suggestion
    -request.user.is_authenticated()
    +requestuser.is_authenticated
    
    -self.request.user.is_anonymous()
    +self.request.user.is_anonymous
```

src/django_upgrade/fixers/request_user_is_authenticated.py
```diff
@@ -0,0 +1,68 @@
...
+
+@fixer.register(ast.Call)
+def visit_Call(
+    state: State,
+    node: ast.Call,
+    parents: list[ast.AST],
+) -> Iterable[tuple[Offset, TokenFunc]]:
+    if (
+        isinstance(node.func, ast.Attribute)
+        and node.func.attr == "is_authenticated"
```
adamchainz, line 37:
```suggestion
        and node.func.attr in ("is_anonymous", "is_authenticated")
```

src/django_upgrade/fixers/request_user_is_authenticated.py
```diff
@@ -0,0 +1,68 @@
...
+        and is_request_user_or_self_request_user(node.func.value)
+        and len(node.args) == 0
+    ):
+        yield (
+            ast_start_offset(node),
+            partial(rewrite_user_is_auth),
```
adamchainz, line 43:
```suggestion
            partial(rewrite_user_attribute),
```

src/django_upgrade/fixers/request_user_is_authenticated.py
```diff
@@ -0,0 +1,68 @@
...
+                isinstance(node.value, ast.Attribute)
+                and isinstance(node.value.value, ast.Name)
+                and node.value.value.id == "self"
+                and node.value.attr == "request"
+            )
+        )
+    )
+
+
+def rewrite_user_is_auth(tokens: list[Token], i: int) -> None:
```
adamchainz, line 63:
```suggestion
def rewrite_user_attribute(tokens: list[Token], i: int) -> None:
```

src/django_upgrade/fixers/request_user_is_authenticated .py
adamchainz: Ah, doing two `del` operations will also break things. After deleting a token, the laters ones shift leftwards in position, decreasing their indexes. Best to delete a range instead - leaving a suggestion.

src/django_upgrade/fixers/request_user_is_authenticated.py
```diff
@@ -0,0 +1,68 @@
...
+    j = find(tokens, i, name=NAME, src="is_authenticated")
+    y = find(tokens, j, name=OP, src="(")
+    z = find(tokens, y, name=OP, src=")")
+    del tokens[z]
+    del tokens[y]
```
adamchainz, line 68:
```suggestion
    del tokens[y:z+1]
```

adamchainz (changes requested):
I read [the release notes section](https://docs.djangoproject.com/en/1.10/releases/1.10/#using-user-is-authenticated-and-user-is-anonymous-as-methods) and noted that `is_anonymous` was also changed. Let’s make the fixer cover both (suggestions below), rename the fixer file to `request_user_attributes.py`, and the tests to `test_request_user_attributes.py`.

I will action these changes now to speed things along.

src/django_upgrade/fixers/request_user_is_authenticated .py
Alurith: I was deleting first the closing parenthesis w/ `del tokens[z]` and the the opening one w/ `del tokens[y]` to not shift positions. Doing as you suggest will remove the space between the parentheses in this case.
```python
def test_spaces_between():
    check_transformed(
        "request . user . is_authenticated ( )",
        "request . user . is_authenticated  ",
        settings,
    )
```
Is that fine?

adamchainz: Yes, that is more consistent IMO. It will remove a comment in this obscure case:

```python
request.user.is_anonymous(  # something
)
```

But I think that’s fine.

adamchainz (approved):

adamchainz:
Thank you for your contribution! Released in 1.16.0.

PR #428, opened 2024-02-12
Run per-file conditions once

adamchainz:
Pushing the per-file conditions from fixer functions up to a per-file pre-pass stage can save many function calls and repeat checks. This yields a small but notable performance improvement.

Benchmarked in two ways.

**First,** using [pytest-profiling](https://pypi.org/project/pytest-profiling/) to measure the test suite:

**Before:** 1574686 function calls (1538837 primitive calls) in 1.308 seconds
**After:** 1569930 function calls (1534037 primitive calls) in 1.297 seconds

A removal of 4756 function calls, which is pretty impressive given the test snippets are small. I don’t read much into the performance improvement here.

**Second,** running on a client project of 50,475 lines of code across 268 files. Each condition was run 10 times with hyperfine.

**Before:** 557.4 ms ± 5.9 ms

**After:** 546.2 ms ± 3.9 ms

About 2% faster.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* Run per-file conditions once, yielding a performance improvement of ~2% measured on a real-world project.
+
 1.16.0 (2024-02-11)
 -------------------
 
--- a/src/django_upgrade/data.py
+++ b/src/django_upgrade/data.py
@@ -101,12 +101,12 @@ def visit(
     settings: Settings,
     filename: str,
 ) -> dict[Offset, list[TokenFunc]]:
-    ast_funcs = get_ast_funcs(settings.target_version)
     initial_state = State(
         settings=settings,
         filename=filename,
         from_imports=defaultdict(set),
     )
+    ast_funcs = get_ast_funcs(initial_state)
 
     nodes: list[tuple[State, ast.AST, ast.AST]] = [(initial_state, tree, tree)]
     parents: list[ast.AST] = [tree]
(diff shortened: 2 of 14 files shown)


UnknownPlatypus:
oh so you did went for the conditional approach, nice!

I'm wondering if the cached_properties on the `State` class are still valuable after that?

Also, another idea I had was to leverage the fact that all these state conditions are mutually exclusive, having one of them return `True` should make every other ones return `False` without evaluating the regex at all.

adamchainz:
> I'm wondering if the cached_properties on the `State` class are still valuable after that?

I think they are because not all fixers use the conditions. Some of them check other things first.

> Also, another idea I had was to leverage the fact that all these state conditions are mutually exclusive, having one of them return True should make every other ones return False without evaluating the regex at all.

I think that will be too minor an improvement to make it worth the extra code. It could also backfire as they’re heuristics, so making the first evaluated one trump the others could lead to unintended false negatives or even non-determinism if fixers run in different orders.

UnknownPlatypus:
Just out of curiosity, what was the hyperfine command you used to benchmark your changes ?

adamchainz:
Replying after some delay: I used `git ls-files -z -- '*.py' | xargs -0 django-upgrade --target-version 5.0`, per the readme. The project uses pre-commit but I wanted to cancel out the overhead of running `pre-commit`.

UnknownPlatypus:
No worries! Was not very familiar with hyperfine at the time but did some research in between, it really is a nice tool!

PR #434, opened 2024-03-11
Try uv for tox runs

adamchainz:


adamchainz:
Looks like it may be a little faster but the variability is so high on GHA maybe not worth it for now.

PR #441, opened 2024-04-07
Remove duplicate variable and integer copy in `tokens.py`

UnknownPlatypus:
Was doing some other work and stumble on the `BRACES` variable which is duplicated in this file.
I also noticed some `j = i` proxies I think we can remove since integers are immutable in python.

Have a nice day
Cheers

### DIFF
--- a/src/django_upgrade/tokens.py
+++ b/src/django_upgrade/tokens.py
@@ -20,10 +20,6 @@ OP = "OP"
 PHYSICAL_NEWLINE = "NL"
 STRING = "STRING"
 
-
-BRACES = {"(": ")", "[": "]", "{": "}"}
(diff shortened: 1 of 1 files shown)


src/django_upgrade/tokens.py
```diff
@@ -21,6 +21,9 @@
 STRING = "STRING"
 
 
+BRACES = {"(": ")", "[": "]", "{": "}"}
+OPENING, CLOSING = frozenset(BRACES), frozenset(BRACES.values())
```
adamchainz, line 25: Let’s keep these under the “More complex mini parsers” comment

adamchainz (commented):
Thanks, that’s a nice little tidy up

adamchainz (approved):

PR #465, opened 2024-06-23
Fix AST node parent tracking

adamchainz:


src/django_upgrade/data.py
```diff
@@ -122,15 +124,12 @@ def visit(
...
-    nodes: list[tuple[State, ast.AST, ast.AST]] = [(initial_state, tree, tree)]
-    parents: list[ast.AST] = [tree]
+    nodes: list[tuple[State, ast.AST, tuple[ast.AST, ...]]] = [
+        (initial_state, tree, ())
+    ]
     ret = defaultdict(list)
     while nodes:
-        state, node, parent = nodes.pop()
-        if len(parents) > 1 and parent == parents[-2]:
-            parents.pop()
```
adamchainz, line 131: The problem clause: a parent would only be popped if iterating up exactly one level. It’s possible to go up two or more levels when leaving iteration of deeper nested nodes, and then the parents would never be popped.

Issue #477, opened 2024-08-08
Django 5.0 fixer for `format_html()` without args/kwargs


adamchainz:
### Description

Deprecated in 5.0 with [this release note](https://docs.djangoproject.com/en/5.0/releases/5.0/#features-deprecated-in-5-0):

> Support for calling format_html() without passing args or kwargs will be removed.

[Ticket #34609](https://code.djangoproject.com/ticket/34609), [my blog post](https://adamj.eu/tech/2023/06/15/format-html/).

We could rewrite certain bad calls, like those using `str.format`:

```diff
 from django.utils.html import format_html

-format_html("<h1>{}</h1>".format(title))
+format_html("<h1>{}</h1>", title)
```

There’s a risk here of breaking code that formatted HTML correctly, because e.g. `html.escape()` was used on variables. But given the many times I’ve seen incorrect code, I think it’s worth that risk, especially now we allow fixers to be disabled.

I think we can try:

* [x] `str.format()` calls - #478
* [ ] ~f-strings~ - too hard
* [ ] `format_html("a string constant")` -> `mark_safe("a string constant")` (maybe?)

adamchainz:
Hmm I think I will leave the `mark_safe` fixer, it's probably best to audit such calls.

PR #478, opened 2024-08-08
Add format_html() fixer

adamchainz:
For #477.

I tried this on a client project with many incorrect `format_html()` calls. I found several examples that the fixer would break, where `format_html()` was used as `mark_safe()`, like:

```python
rows = format_html(...)
table = format_html(
    """
    <table>
        <thead>...</thead>
        <tbody>
            {}
        </tbody>
    </table>
    """.format(
        rows
    )
)
```

The fixer would break this, making `rows` get HTML-escaped.

Since it had more breakage than I expected, I am going to let this one sit. Perhaps it’s not worth adding in the end.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog
 =========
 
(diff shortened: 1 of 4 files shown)


adamchainz:
I finished auditing that project and found there weren’t so many such instances, in the end. I just bumped into them at the start. I think it’s worth adding this fixer.

PR #482, opened 2024-08-14
Extend format_html() fixer to string constants

adamchainz:
For #477.


adamchainz:
Started with the easy case, when there is an existing `mark_safe` import. I want to extend this to add the import if it’s missing, and the name `mark_safe` is not used anywhere in the file...

adamchainz:
Giving up on this idea, best to audit such calls any way.

PR #484, opened 2024-08-19
Add gis_admin_deprecations fixer

fitoria:
Allows replacement of OSMGeoAdmin and GeoModelAdmin with GISModelAdmin from the same module, those classe were removed on Django 5.0: https://docs.djangoproject.com/en/5.0/releases/5.0/#django-contrib-gis

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,6 +21,10 @@ Pending
 
   `PR #660 <https://github.com/adamchainz/django-upgrade/pull/660>`__.
 
+* Add Django 4.0+ fixer :ref:`gis_admin_deprecations <gis_admin_deprecations>` to migrate from deprecated GIS admin classes to the new ones.
+
+  Thanks to Adolfo Fitoria in `PR #484 <https://github.com/adamchainz/django-upgrade/pull/484>`__.
+
 * Extend :ref:`test_http_headers <test_http_headers>` fixer to cover ``AsyncClient``, ``AsyncRequestFactory``, and ``self.async_client.*()`` calls.
 
   Thanks to Benjamin Aduo in `PR #633 <https://github.com/adamchainz/django-upgrade/pull/633>`__.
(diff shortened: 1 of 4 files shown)


adamchainz:
Oh great. I haven't got time for a full review right now, but at a glance it’s looking good. Can you add a changelog note?

Little note for the future - please don't make PRs from the `main` branch of your fork, but create a branch. It makes it easier for maintainers to push edits back and be sure that things don't break.

fitoria:
Ready!

adamchainz:
This would be a Django 4.0 fixer, because that's when the classes were deprecated. I have moved it there.

I looked at the code as of Django 4.0. It looks like we can't blindly rewrite model admin classes to use `GISModelAdmin` in the case that they define any of the 29 attributes provided by the old classes: https://github.com/django/django/blob/4555aa0a489cb9dcf764edf12339097cdfa5ff84/django/contrib/gis/admin/options.py#L43-L71 . Doing so could break functionality.

Those attributes *could* also come from multiple inheritance, which should also mean that a safe automated fix is impossible.

So I think we should add two restrictions:

1. Only rewrite model admin classes that inherit from exactly one class, with no keyword arguments (i.e. not `class F(OSMGeoAdmin, something=1):`, just in case)
2. Ignore any classes that have any the 29 attributes defined in their class bodies.

Do you think you could update to add these, and appropriate tests?

fitoria:
> Do you think you could update to add these, and appropriate tests?

Sure, I'm going to take care of it, thanks for the review.

adamchainz:
Hey, I am just gearing up for a release here, so I took a look at your PR. It needs quite a bit of work to limit the fixer to class definitions and add the above restrictions. I pushed two new, failing tests that represent the restrictions, and will leave the PR like this for now. Let me know if you find time to work on this again (no pressure!).

adamchainz:
I've just rebased and pushed a new commit that addresses my review comments. However, I'm coming to think with all the exclusions added here and the niche nature of GIS admin, it may not be worth actually adding this fixer. I'm gonna ponder it and come back to make a decision.

adamchainz:
Yeah, I think we won't add this one. Thanks for your work here @fitoria , it helped show how hard the "simple" replacement is!

PR #487, opened 2024-09-05
Use uv on GitHub Actions

adamchainz:


adamchainz:
After testing here, I have a script to run against all my repos, doing that now.

Issue #495, opened 2024-10-10
Unrelated comment was deleted after a change


BonaFideIT:
### Python Version

3.11.10

### Django Version

4.2.15

### Package Version

1.21.0

### Description

Hi,

while upgrading the `devel` branch from https://github.com/django-guardian/django-guardian to a new django version, we run django-upgrade like this `git ls-files -z -- '*.py' | xargs -0 django-upgrade --target-version 5.0`.

The following code of https://github.com/django-guardian/django-guardian/blob/devel/guardian/__init__.py:

```python
from . import checks
import django

if django.VERSION < (3, 2):
    default_app_config = 'guardian.apps.GuardianConfig'

# PEP 396: The __version__ attribute's value SHOULD be a string.
__version__ = '2.4.0'

...
```

changed to:
```python
from . import checks
import django

__version__ = '2.4.0'

...
```

We think the comment `# PEP 396: The __version__ attribute's value SHOULD be a string.` should not be removed.

PR #496, opened 2024-10-10
Avoid accidental removal of trailing comments

BonaFideIT:
We added a test case for the issue mentioned in https://github.com/adamchainz/django-upgrade/issues/495.


adamchainz:
Thank you for the failing test case, that really helped me nail the fix. Merging now and releasing shortly.

PR #517, opened 2024-12-15
Support single tuple versioned block fixer

### LINKED ISSUE #509: Extend version block remover to single tuples
### Description

I had one of these in django-rich:

```python
if django.VERSION >= (4,):
    ...
```

UnknownPlatypus:
Fixes #509 

I intentionally skipped 0-length tuple because it's most likely not intentional and removing code based on that would be surprising I think.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,10 @@ Changelog
 
 * Support Django 5.2 in ``--target-version``.
 
+* Extend versioned block fixer to support comparisons with single-item tuples, like `django.VERSION >= (4,)`.
+
+  Thanks to Thibaut Decombe in `PR #517 <https://github.com/adamchainz/django-upgrade/pull/517>`__.
+
 1.22.2 (2024-12-02)
 -------------------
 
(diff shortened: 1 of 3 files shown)


src/django_upgrade/ast.py
```diff
@@ -74,7 +74,7 @@ def is_passing_comparison(
         and isinstance(test.ops[0], (ast.Gt, ast.GtE, ast.Lt, ast.LtE))
         and len(test.comparators) == 1
         and isinstance((comparator := test.comparators[0]), ast.Tuple)
-        and len(comparator.elts) == 2
+        and 0 < len(comparator.elts) <= 2
```
adamchainz, line 77: A little more obvious
```suggestion
        and 1 <= len(comparator.elts) <= 2
```

CHANGELOG.rst
```diff
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+* Support single-tuple Django version in the versioned block fixer.
```
adamchainz, line 5:
```suggestion
* Extend versioned block fixer to support comparisons with single-item tuples, like `django.VERSION >= (4,)`.
```

adamchainz (changes requested):

adamchainz:
> I intentionally skipped 0-length tuple because it's most likely not intentional and removing code based on that would be surprising I think.

Indeed!

adamchainz (approved):

PR #542, opened 2025-03-27
Restore Django 5.0+ model_field_choices fixer

### LINKED ISSUE #417: `.choices` dropped for choices not subclassing the `*Choices` class
### Python Version

3.12.1

### Django Version

5.0.1

### Package Version

1.15.0

### Description

I have a number of choice classes in a parent-child relationship. Such a relationship is the reason why, in those classes, I am not inheriting from any Django `*Choices` class.

I am using these choice classes in `choices` field params and `django-upgrade` appears to be dropping the `.choices` attribute since it is supporting Django 5.0. Passing that attribute is vital for such choice classes to work, though.

Shouldn't `django-upgrade` detect whether the passed `.choices` belongs to a class inheriting from any Django `*Choices` class, and only in that case drop the `.choices` attribute?

adamchainz:
Fixes #417.

Restore the fixer with a new restriction that we only act when the choices class is defined in the same file and can be reliably detected as subclassing one of Django's enumeration types.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,10 @@ Changelog
 
 * Add Django 5.2+ fixer ``staticfiles_find_all`` to rewrite calls to the staticfiles ``find()`` function using the old argument name ``all`` to the new name ``find_all``.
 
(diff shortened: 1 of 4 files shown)


UnknownPlatypus:
Thanks for this adam :raised_hands:

Issue #548, opened 2025-04-04
Support Class-Based Settings


gav-fyi:
### Description

Hi. 👋🏻 

I'm not sure how niche this is, but wanted to suggest a new feature of supporting class based settings files. Like those you get with [django-classy-settings](https://django-classy-settings.readthedocs.io/latest/).

Example:

```python
# project/settings/base.py
class BaseSettings:

    INSTALLED_APPS = [...]

    USE_L10N = True  # should be removed
```

Currently, `django-upgrade` would ignore these settings as they aren't top-level. 

If this is something you're willing to take on, I'm happy to raise a PR for the changes to the current settings fixers. 

Cheers 

---

### Todo

- Fixers:
  - [x] `USE_L10N` ([PR](https://github.com/adamchainz/django-upgrade/pull/549))
  - [x] `PASSWORD_RESET_TIMEOUT_DAYS` ([PR](https://github.com/adamchainz/django-upgrade/pull/551))
  - [x] `django.db.backends.postgresql_psycopg2` in `DATABASES` ([PR](https://github.com/adamchainz/django-upgrade/pull/552))
  - [x] `STORAGES` ([PR](https://github.com/adamchainz/django-upgrade/pull/553))

adamchainz:
I think that’s acceptable, yeah, let’s do it. Please make a checklist of all the fixers and do one PR for each!

adamchainz:
Thanks @gav-fyi . I aimed to review and merge your work at DjangoCon Europe, when I also migrated the project docs to Sphinx and added the logo, but ran out of time. I found some time just now, though.

I rebased and merged three PRs, but closed #553. I left a comment there on why I think it's not a safe fix to make.

Thank you for your work!

Also, FYI, there seemed to be a mixup between the email addresses you used on GitHub and your local Git, for d1c0838e8bba61c881c25b75a8b176ebf11f503c .

gav-fyi:
Thanks for the review. 👍🏻 

On reflection, I think closing #553 is the safest option, as I've also thought of a couple of other inheritance-based configurations that it will likely not cover too.

> Also, FYI, there seemed to be a mixup between the email addresses you used on GitHub and your local Git, for https://github.com/adamchainz/django-upgrade/commit/d1c0838e8bba61c881c25b75a8b176ebf11f503c .

My bad. I must have pushed that from the wrong profile.

PR #553, opened 2025-04-08
Add support for class-based settings files in STORAGES fixer

gav-fyi:
Add support for class-based settings files for the `STORAGES` settings fixer.

Partially Implements #548

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* Add support for class-based settings files in ``STORAGES`` fixer.
+
 1.24.0 (2025-03-27)
 -------------------
 
--- a/README.rst
+++ b/README.rst
@@ -405,6 +405,8 @@ Only applies if all old settings are defined as strings, at module level, and a 
 Settings files are heuristically detected as modules with the whole word “settings” somewhere in their path.
 For example ``myproject/settings.py`` or ``myproject/settings/production.py``.
 
+Class-based settings files, such as ``django-configurations`` and ``django-classy-settings``, are supported.
+
 .. code-block:: diff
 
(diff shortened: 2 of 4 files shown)


adamchainz:
Thanks for the PR here. I think, though, the fix is not so safe to operate on within classes, and in the spirit of caution, I think we shouldn’t merge this PR.

I’m worried about situations where several classes are mixed-in to a single settings class, like:

```py
class CoreSettings:
    DEFAULT_FILE_STORAGE = "example.backend"
    
class StaticSettings:
    STATICFILES_STORAGE = "example.backend"
    
class ProductionSettings(CoreSettings, StaticSettings):
    pass
```

Currently, the fixer is broken for this situation, emptying `StaticSettings` and moving its declaration into `CoreSettings`:

```py
 class CoreSettings:

    STORAGES = {
        "default": {
            "BACKEND": "example.backend",
        },
        "staticfiles": {
            "BACKEND": "example.backend",
        },
    }
 class StaticSettings:

 class ProductionSettings(CoreSettings, StaticSettings):
     pass
```

But even if that was fixed, it would have to put `STORAGES` in both classes:

```py
class CoreSettings:
    STORAGES = {
        "default": {
            "BACKEND": "example.backend",
        },
    }
    
 class StaticSettings:
    STORAGES = {
        "default": {
            "BACKEND": "django.core.files.storage.FileSystemStorage",
         },
         "staticfiles": {
            "BACKEND": "example.backend",
         },
     }

 class ProductionSettings(CoreSettings, StaticSettings):
     pass
```


The, the inheritance in `ProductionSettings` will inherit just the `STORAGES` value from `CoreSettings`, losing the staticfiles backend, and breaking the project. And since Django’s warning would be fixed, it would be harder to debug than the original situation. I believe mixing settings classes together is quite common with class-based settings libraries, so I think we should avoid this kind of fix.

PR #564, opened 2025-05-15
Freeze pre-commit hooks

adamchainz:
Doing so prevents surprise updates and stops pre-commit warning about mutable tags, such as on the crate-ci/typos repo.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ default_language_version:
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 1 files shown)


UnknownPlatypus:
Do you know if pre-commit.ci understand that it now has to run with the `--freeze` flag ? 
I did not found [configuration](https://pre-commit.ci/#configuration) about it

adamchainz:
It detects frozen commits and adds `--freeze` automatically, but I could only find this issue thread as a reference: https://github.com/pre-commit-ci/issues/issues/221.

PR #618, opened 2025-12-09
Swap BadHeaderError to ValueError. Its deprecated.

awais786:
https://github.com/adamchainz/django-upgrade/issues/616

### DIFF
--- a/docs/fixers.rst
+++ b/docs/fixers.rst
@@ -260,6 +260,31 @@ Django 5.1
 
 `Release Notes <https://docs.djangoproject.com/en/5.1/releases/5.1/>`__
 
+``BadHeaderError``
+~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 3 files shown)


adamchainz:
This change was in Django 6.0: https://docs.djangoproject.com/en/6.0/releases/6.0/#:~:text=django%2Ecore%2Email%2EBadHeaderError (not Django 5.1 per the docstring in the fixer here).

I don't think we can safely replace the exception class here. `ValueError` is incredibly common, so code like this:

```diff
 try:
     send_email()
     do_other_thing_that_might_raise_value_error()
-except BadHeaderError:
+except ValueError:
     pass
 ```
 
…would suddenly start passing when `do_other_thing_that_might_raise_value_error` raises, rather than allowing it to propagate.

So thanks for taking a look at contributing, but this PR won't be merged. Thanks!

Issue #389, opened 2023-09-24
Add `@admin.display` decorators in models file too


UnknownPlatypus:
### Description

I think the display fixer dosn't support this use case [from the docs](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display): 

- A string representing a model attribute or method (without any required arguments).

Would be nice to have it work in that case too: 

```diff
+from django.contrib import admin
from django.db import models

class Person(models.Model):
    name = models.CharField(max_length=50)
    birthday = models.DateField()

+   @admin.display(description='Birth decade')
    def decade_born_in(self):
        return '%d’s' % (self.birthday.year // 10 * 10)
-   decade_born_in.short_description = 'Birth decade'
```

adamchainz:
Yeah good point, would be nice to fix that.

UnknownPlatypus:
I'm not quite sure how to gracefully handle adding the `from django.contrib import admin` import if missing. Maybe you have an idea how to do it ?

adamchainz:
I have ideas but haven’t implemented anything. I gave up on adding imports in the [timezone.utc fixer](https://github.com/adamchainz/django-upgrade#djangoutilstimezoneutc-deprecations).

Something like “add after the last import statement at the top of the file” could work. We’d probably want to check that there’s nothing called `admin` in the module...

UnknownPlatypus:
> add after the last import statement at the top of the file
This seems manageable, finding the last import in the module is not too hard. Doing that we can also check for possible import with the same name

> We’d probably want to check that there’s nothing called admin in the module...

I feel like this cost a lot, we would have to check every import, every variable assignment, every function/class definition and if there is any wildcard import, this is unsafe (but I'd say we can ignore this case). I'm not sure what do do here

adamchainz:
I agree it would be a bit costly, but we’d only need to do it in the case where we’d like to add the import. Anyway, this particularly fixer feels somewhat niche, at least in my experience I haven’t seen admin methods on models.

PR #631, opened 2026-03-09
Extend `utils_timezone` fixer work when no datetime import exists

### LINKED ISSUE #568: utils_timezone fixer doesn't seem to consider `ast.ImportFrom` nodes.
### Python Version

3.10.11

### Django Version

4.2.21

### Package Version

1.25.0

### Description

Hi Adam 👋 

When trying to apply the `utils_timezone` fixer to some files I noticed it failed to apply some fixes. Below is a simple set of code you can use to reproduce:

```python
from django.utils import timezone
from datetime import datetime

print(datetime.now(tz=timezone.utc))
```

Running `django-upgrade --target-version=5.2 --only=utils_timezone simple_reproduction.py` will not apply any transforms. At a quick glance it seems like when building the `ImportDetails` the code [isn't looking](https://github.com/adamchainz/django-upgrade/blob/1.25.0/src/django_upgrade/fixers/utils_timezone.py#L104-L110) for `ast.ImportFrom` nodes when identifying the `datetime_module`.

Thanks for your work on this package!

benaduo:
## Summary

The `utils_timezone` fixer silently skipped files that had no `import datetime`
 statement, even when `django.utils.timezone.utc` or bare `utc` references were present and eligible for rewriting. This affected any codebase using
 `from datetime import ...` style imports alongside the deprecated `django.utils.timezone.utc`.

## Motivation

Fixes #568. The root cause was that `get_import_details()` only scanned for `ast.Import` nodes to set `datetime_module`, ignoring `ast.ImportFrom` nodes entirely. As suggested in the issue, the fix inserts `import datetime as dt` when `dt` is a free name in the module, so existing code patterns like `from datetime import datetime` no longer block the transformation.

**Before:**

```
python # Input — fixer produced no output

from datetime import datetime
from django.utils import timezone

print(datetime.now(tz=timezone.utc))
```

**After:**

``` 
import datetime as dt
from datetime import datetime
from django.utils import timezone

print(datetime.now(tz=dt.timezone.utc))
```

**Changes**

`src/django_upgrade/fixers/utils_timezone.py`

Add `_is_name_used():` walks the full module AST to check whether a name is already bound anywhere, Name nodes, import aliases, function/class definitions, and function arguments
Extend `ImportDetails` with `first_import_node`, `needs_datetime_import`, and `add_import_scheduled` fields
Extend `get_import_details()` to fall back to `dt` as the `datetime` alias when no `import datetime` is found and dt is a free name
Add `add_datetime_import()` token function to insert import datetime as dt before the first import in the file
Add `maybe_add_datetime_import()` generator, mirroring `maybe_rewrite_import()`
Call `maybe_add_datetime_import()` from both `visit_Name` and `visit_Attribute`

`tests/fixers/test_utils_timezone.py`

Update `test_no_datetime_import` from `check_noop` to `check_transformed`
Add `test_no_datetime_import_dt_name_conflict: noop` when dt is already used
Add `test_from_datetime_import_attr:` from `datetime import datetime` + `timezone.utc` → inserts import, rewrites
Add `test_from_datetime_import_name:` from `datetime import datetime` + bare `utc` → inserts import, removes old import, rewrites

**Testing**
20/20 utils_timezone fixer tests pass
981/981 full test suite passes

**Author**
Benjamin Aduo ([@benaduo](https://github.com/benaduo))

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,10 @@ Pending
 
   Thanks to Benjamin Aduo in `PR #633 <https://github.com/adamchainz/django-upgrade/pull/633>`__.
 
+* Extend :ref:`utils_timezone` fixer to work when no ``import datetime`` statement exists, by inserting ``import datetime as dt`` when the name ``dt`` is unused.
+
+  Thanks to Ryan Siemens for the report in `Issue #568 <https://github.com/adamchainz/django-upgrade/issues/568>`__ and Benjamin Aduo for the implementation in `PR #631 <https://github.com/adamchainz/django-upgrade/pull/631>`__.
+
 * Extend :ref:`versioned_test_skip_decorators <versioned_test_skip_decorators>` fixer to also remove the entire decorated function or class when the skip condition is always true.
   For example, when targeting Django 5.2+, a function decorated with ``@pytest.mark.skipif(django.VERSION >= (5, 2), ...)`` will be removed.
 
--- a/docs/fixers.rst
+++ b/docs/fixers.rst
@@ -471,6 +471,8 @@ Django 4.1
 
 `Release Notes <https://docs.djangoproject.com/en/4.1/releases/4.1/>`__
 
+.. _utils_timezone:
+
 ``django.utils.timezone.utc`` deprecations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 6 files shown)


adamchainz:
Thanks for this PR too. The 

I've made more edits here too: docs updates, changelog link, extra tests, and moving `_is_name_used` to a different, cached utility `get_module_names`. This will mean only one extra pass, when needed, and we can share the data between fixers as required (e.g. #389 would like this data).

adamchainz:
I found a bug and fixed it in the last commit!

PR #647, opened 2026-06-14
Add Django 1.11+ fixer for @models.permalink

### LINKED ISSUE #645: Support @permalink cleanup
### Description

One of the most annoying conversions in old Django code bases is the removal of `django.db.models.permalink` in Django 2. It would be nice to have a fixer for simple cases, e.g. functions that just return a tuple.

adamchainz:
Fixes #645.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,11 @@ Pending
 
   `PR #649 <https://github.com/adamchainz/django-upgrade/pull/649>`__.
 
+* Add Django 1.11+ fixer :ref:`permalink <permalink>` to rewrite ``@models.permalink`` decorator usages to use ``reverse()`` directly.
+
+  `PR #647 <https://github.com/adamchainz/django-upgrade/pull/647>`__.
+  Thanks to Joerg Sonnenberger for the report in `Issue #645 <https://github.com/adamchainz/django-upgrade/issues/645>`__ and review.
+
 1.30.0 (2026-02-24)
 -------------------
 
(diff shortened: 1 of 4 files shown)


jsonn:
Looks good and covers over 90% of my old project. What is not covered:

- combination of `@classmethod`  and `@permalink`, even if the latter is the inner decorator?
- decorated methods in `forms.py`
- non-trivial methods e.g. with conditionals and multiple returns.

A small Quality of Implementation issue: if the `args` tuple or the `kwargs` dict is empty, shouldn't they be skipped?

Thanks!

adamchainz:
Thanks for the feedback and helping get this mergeable.

> Looks good and covers over 90% of my old project.

So, to be clear, you ran it?

> combination of `@classmethod` and `@permalink`, even if the latter is the inner decorator?

I don't think this combination was ever intended to be supported, so I don't think we should support rewriting it.

> decorated methods in `forms.py`

Afaiu `@permalink` was intended for model methods, so usage in `forms.py` files seems like it should be rare, too.

> non-trivial methods e.g. with conditionals and multiple returns.

This is hard to get right, so I don't believe it's worth the investment.

> A small Quality of Implementation issue: if the `args` tuple or the `kwargs` dict is empty, shouldn't they be skipped?

Good spot, added!

jsonn:
Yes, I tested it and checked the diff.  After that, I checked what remained. I don't believe any of those are big, the classmethod case was the biggest issue left in my own code, but I agree that it might not be idiomatic.

adamchainz:
Great, thanks for the quick reply. Gonna merge this now!

Issue #654, opened 2026-06-22
Django 6.1: merge `EMAIL_*` settings into `MAILERS`


adamchainz:
### Description

Per [release note](https://docs.djangoproject.com/en/dev/releases/6.1/#mailers-deprecations):

> The `EMAIL_BACKEND`, `EMAIL_FILE_PATH`, `EMAIL_HOST`, `EMAIL_HOST_PASSWORD`, `EMAIL_HOST_USER`, `EMAIL_PORT`, `EMAIL_USE_TLS`, `EMAIL_USE_SSL`, `EMAIL_SSL_CERTFILE`, `EMAIL_SSL_KEYFILE`, and `EMAIL_TIMEOUT` settings are deprecated. Replace them with a [`MAILERS`](https://docs.djangoproject.com/en/dev/ref/settings/#std-setting-MAILERS) configuration dictionary.

…we can merge these settings into `MAILERS`, like we did for `DEFAULT_FILE_STORAGE`/`STATICFILES_STORAGE` → `STORAGES` in 4.2.

medmunds:
There's a potential compatibility concern with this raised in [DEP 0018](https://github.com/django/deps/blob/main/accepted/0018-mailers.md#django-upgrade-recommendations:~:text=(This%20example%20suggests,non%2Ddefault%20backends.)).

adamchainz:
Right, thanks for writing that back in the DEP. Cheers!

adamchainz:
Hmm, per this note:

> This example suggests django-upgrade may not ever be able to safely upgrade existing settings,

…and the limited scope:

> projects that use only the default SMTP EmailBackend 

I think we won't take any action here. It seems like so much depends on how ready third-party backends are, etc., that we should leave developers to handle the upgrade themselves. It's possible that popular third-party backends will provide their own migration guides, too (?), in which case users will at least have plenty of guidance.

Thanks for the clear write-up, @medmunds !

ulgens:
Is there a way to implement "unsafe" fixes and/or "warning only" messages? I mean, I understand the reasoning here, auto updating them is considered unsafe, but I'd still expect django-upgrade to warn me about these changes - and I'd be okay with the tool applying these changes after I approved that I understand they are unsafe.

adamchainz:
There's no reason to implement warnings in django-upgrade as Django itself will warn you.

I don't think an "unsafe" fixer here would serve anyone well. It would only leave you with almost-certainly broken settings.

PR #662, opened 2026-06-23
Add Django 6.1 fixer for `AdminEmailHandler` `email_backend` argument rename.

### LINKED ISSUE #652: Django 6.1: rename `email_backend` argument of `AdminEmailHandler` to `using`
### Description

Per [release note](https://docs.djangoproject.com/en/dev/releases/6.1/#:~:text=email_backend%20argument):

> The `email_backend` argument of [`AdminEmailHandler`](https://docs.djangoproject.com/en/dev/topics/logging/#django.utils.log.AdminEmailHandler) is deprecated in favor of the newly introduced `using` argument.

…we can rename the keyword argument, like we did for `ordering` → `order_by` in 5.2.

adamchainz:
Fixes #652.

### DIFF
--- a/docs/fixers.rst
+++ b/docs/fixers.rst
@@ -132,6 +132,30 @@ Django 6.1
 
 `Release Notes <https://docs.djangoproject.com/en/6.1/releases/6.1/>`__
 
+.. _settings_logging_admin_email_handler:
+
+``AdminEmailHandler`` ``email_backend`` argument
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Name:** ``settings_logging_admin_email_handler``
+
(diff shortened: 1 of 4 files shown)


docs/fixers.rst
```diff
@@ -132,6 +132,30 @@ Django 6.1
...
+For example ``myproject/settings.py`` or ``myproject/settings/production.py``.
+
+.. code-block:: diff
+
+    LOGGING = {
+        "handlers": {
+            "mail_admins": {
+                "class": "django.utils.log.AdminEmailHandler",
+   -            "email_backend": "myapp.backends.EmailBackend",
+   +            "using": "myapp.backends.EmailBackend",
```
medmunds, line 154: @adamchainz this won't work. `"using"` is an *alias* to a `MAILERS` configuration (a key in the `MAILERS` dict—e.g., `"default"` or `"notifications"` or `"admin-log-stuff"`), **not** an email backend import path.

adamchainz: Right you are, thanks for checking. I will remove this fixer since there's no way to reliably add the `MAILERS` setting, as discussed in #654.

PR #674, opened 2026-07-15
Add Django 7.0 fixer for USE_BLANK_CHOICE_DASH setting removal.

benaduo:
Removes the transitional \USE_BLANK_CHOICE_DASH = True\ setting from settings files, as it no longer has any effect since Django 7.0.

Operates on module-level and class-level assignments in heuristically-detected settings files.

Ref: https://docs.djangoproject.com/en/dev/releases/7.0/#features-removed-in-7-0

### DIFF
--- a/docs/fixers.rst
+++ b/docs/fixers.rst
@@ -220,6 +220,27 @@ Renames the deprecated ``savepoint`` function in ``django.db.transaction`` to ``
     +from django.db import transaction
     +transaction.savepoint_create("name")
 
+Django 7.0
+----------
+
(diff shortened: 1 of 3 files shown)


adamchainz:
Django 7.0 isn't even under development and [will be called Django 2028](https://www.djangoproject.com/weblog/2026/aug/10/annual-release-cycle/). I'm gonna park this in draft until it's underway, maybe when it's out as an alpha.

adamchainz:
@benaduo there is no need to keep this branch up to date.

